PR #140, opened 2013-09-02
Math.roundFloat32

ljharb:
Adding Math.roundFloat32.

There's a good bit of code, from https://github.com/inexorabletash/polyfill/blob/master/typedarray.js#L172-L278 (with permission), to support proper number truncation in an environment without TypedArrays.

There's also a TypedArray solution that will be much faster in newer environments.

### DIFF
--- a/es6-shim.js
+++ b/es6-shim.js
@@ -821,7 +821,7 @@
         return ((al * bl) + (((ah * bl + al * bh) << 16) >>> 0)|0);
       },
 
-      roundFloat32: function(x) {
+      fround: function(x) {
         if (x === 0 || x === Infinity || x === -Infinity || Number.isNaN(x)) {
           return x;
         }
--- a/test/math.js
+++ b/test/math.js
@@ -318,55 +318,55 @@ describe('Math', function() {
     });
   });
 
-  describe('Math.roundFloat32', function() {
+  describe('Math.fround', function() {
     // Mozilla's reference tests: https://bug900125.bugzilla.mozilla.org/attachment.cgi?id=793163
     it('returns NaN for undefined', function() {
-      expect(Number.isNaN(Math.roundFloat32())).to.be.ok;
+      expect(Number.isNaN(Math.fround())).to.be.ok;
     });
 
     it('returns NaN for NaN', function() {
-      expect(Number.isNaN(Math.roundFloat32(NaN))).to.be.ok;
+      expect(Number.isNaN(Math.fround(NaN))).to.be.ok;
     });
 
     it('works for zeroes and infinities', function() {
-      expect(isPositiveZero(Math.roundFloat32(0))).to.be.ok;
-      expect(isNegativeZero(Math.roundFloat32(-0))).to.be.ok;
-      expect(Math.roundFloat32(Infinity)).to.equal(Infinity);
-      expect(Math.roundFloat32(-Infinity)).to.equal(-Infinity);
+      expect(isPositiveZero(Math.fround(0))).to.be.ok;
+      expect(isNegativeZero(Math.fround(-0))).to.be.ok;
+      expect(Math.fround(Infinity)).to.equal(Infinity);
+      expect(Math.fround(-Infinity)).to.equal(-Infinity);
     });
 
     it('returns infinity for large numbers', function() {
-      expect(Math.roundFloat32(1.7976931348623157e+308)).to.equal(Infinity);
-      expect(Math.roundFloat32(-1.7976931348623157e+308)).to.equal(-Infinity);
-      expect(Math.roundFloat32(3.4028235677973366e+38)).to.equal(Infinity);
+      expect(Math.fround(1.7976931348623157e+308)).to.equal(Infinity);
+      expect(Math.fround(-1.7976931348623157e+308)).to.equal(-Infinity);
+      expect(Math.fround(3.4028235677973366e+38)).to.equal(Infinity);
     });
 
     it('returns zero for really small numbers', function() {
       expect(Number.MIN_VALUE).to.equal(Math.pow(2, -1074)); // sanity check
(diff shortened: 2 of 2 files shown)


Yaffle:
Seems, this implementation is a little buggy.
Try to use `Math.fround(1.9999999999999998)` in a browser without Float32Array support and without native `Math.fround`.

P.S.
I think, you might remove `Math.fround` from es6-shim, Typed Arrays are part of the ES6 spec too, and polyfill for them, may include `Math.fround`.

ljharb:
@Yaffle I'm not sure what you mean. The `fround` shim works perfectly fine with that value, and `Float32Array` isn't used whatsoever. Why would we remove a perfectly working shim?

Yaffle:
@ljharb , really?

Math.fround(1.9999999999999998) !== 2 for me

Yaffle:
@ljharb , you can remove 150+ lines of buggy and useless code

ljharb:
@Yaffle In which browser are you testing this?

It's not useless if it works in all but a few edge cases, and if it was buggy, we'd have tests indicating it :-) if you've found a bug, then let's make a test case and fix it. It will *never* be removed as long as it's possible to keep it compliant. "less lines of code" is never a useful goal.

Yaffle:
@ljharb , `delete Float32Array; delete Math.fround; <script src="es6-shim.js"></script><script>alert(Math.fround(1.9999999999999998).toString())</script>`

ljharb:
Ah, I forgot I used `Float32Array` in there. OK, thanks, i'll see if I can come up with a test.

Although, do you have a specific browser I can test this in that already lacks Typed Arrays and `fround`?

Yaffle:
@ljharb , `float` can store `2 * (1 - Math.pow(2, -24) === 1.9999998807907104` and `2`, the last value i s close to the `1.9999999999999998`

ljharb:
ah, so it's about breaking a tie between whether `fround` produces `1.9999998807907104` or `2`, for `1.9999999999999998`?

You've submitted lots of helpful PRs and issues about `Math` function tiebreaking - perhaps we can do the same here :-)

Yaffle:
@ljharb , yes, my explanation make no sense, because buggy `Math.fround(1.9999999999999998)` returns value bigger than 2.

I have no use cases, just do not want to see buggy code, and as `Math.fround` can be implemented in ~5 lines of code using `Float32Array`,  people might use polyfill for typed arrays - https://github.com/inexorabletash/polyfill/blob/master/typedarray.js and might create an issue about this bug in that repository.

ljharb:
If they use a polyfill for `Float32Array` then `Math.fround` will pick it up and use that (and as you pointed out, a bug in that polyfill is that polyfill's problem) - the issue would be when *not* using a polyfill, in a browser without Typed Arrays or `Math.fround`. I'd still like to fix it, if we can figure out how.

Yaffle:
I can propose next implementation:

```

    function roundTiesToEven(n) {
      return (n + 1 / Number.EPSILON) - 1 / Number.EPSILON;
    }

    function fround(v) {
      var EPSILON = Math.pow(2, -23);
      var MAX_VALUE = Math.pow(2, 127) * (2 - EPSILON);
      var MIN_EXPONENT = -126;

      if (v !== v) {
        return v;
      } else if (v === Infinity || v === -Infinity) {
        return v;
      } else {
        var sign = Math.sign(v);
        var abs = Math.abs(v);

        var e = Math.floor(Math.log(abs) * Math.LOG2E + 0.5);
        var t = Math.pow(2, e);
        if (t > abs) {
          e -= 1;
        }
        if (t * 2 <= abs) {
          e += 1;
        }

        if (e < MIN_EXPONENT) {
          e = MIN_EXPONENT;
        }
        t = Math.pow(2, e);
        var result = roundTiesToEven(abs / t / EPSILON) * t * EPSILON;
        if (result > MAX_VALUE) {
          result = Infinity;
        }
        return sign * result;
      }
    }

```

Yaffle:
@ljharb 

IMHO `fround` is useless, but I can propose next code:

```
    function roundTiesToEven(n) {
      return (n + 1 / Number.EPSILON) - 1 / Number.EPSILON;
    }

    function fround(v) {
      var EPSILON = Math.pow(2, -23);
      var MAX_VALUE = Math.pow(2, 127) * (2 - EPSILON);
      var MIN_VALUE = Math.pow(2, -126);
      var abs = Math.abs(v);
      var sign = Math.sign(v);
      if (abs < MIN_VALUE) {
        return sign * roundTiesToEven(abs / MIN_VALUE / EPSILON) * MIN_VALUE * EPSILON;
      }
      // Veltkamp's splitting (?)
      var a = (1 + EPSILON / Number.EPSILON) * abs;
      var result = a - (a - abs);
      if (result > MAX_VALUE || result !== result) {
        return sign * Infinity;
      }
      return sign * result;
    }
```
Do not ask me how it works - I do not know.

ljharb:
Um - how does `roundTiesToEven` work? That just reduces down to `return n`

Yaffle:
@ljharb , do not think so, `roundTiesToEven` works same as `roundToEven`  https://github.com/paulmillr/es6-shim/blob/master/es6-shim.js#L389

ljharb:
Clever. I'll update the `fround` implementation now in a PR.

zloirock:
@Yaffle wow... cool :)

Yaffle:
@zloirock , thanks

Yaffle:
The `fround` implementation - https://github.com/paulmillr/es6-shim/pull/140#issuecomment-91787165 - is incorrect.
Native method gives +-0 on some devices with Mobile Safari:
Math.fround(-1.1754943508222874e-38) === -0
Math.fround(+1.1754943508222874e-38) === +0
May be, denormalized values should be flushed to zero for "binary32" too...

PR #147, opened 2013-10-03
Don't require descriptor support to implement Map and Set. If descriptors are not supported, then the 'size' property is overwritten directly rather than proxying to '_size'.

ckknight:
Technically this would mean that size is overwritable in non-ES5 implementations, but aside from that, it works as expected.

### DIFF
--- a/es6-shim.js
+++ b/es6-shim.js
@@ -823,11 +823,11 @@
           }
         }
 
-        function updateSize(func) {
+        var updateSize = function(func) {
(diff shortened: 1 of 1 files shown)


ljharb:
If you wanted to add this into the `es6-sham` at some point, I'd actually be open to it - provided that the tests for `es6-shim` enforced an unsettable `size`, and the `es6-sham` tests didn't.

PR #313, opened 2014-12-29
Reflect

Victorystick:
The Reflect global is, as far as I know, a part of the ES6 spec. I was a bit surprised not to find it in the es6-shim. Is there any reason as to why it isn't? If not, I'd consider giving it a go.

I'd like to use this PR to discuss Reflect implementation.


es6-shim.js
```diff
@@ -2097,6 +2097,88 @@
...
+            throw new TypeError('target must be an object');
+          }
+
+          var keys = [];
+
+          for (var key in target) {
+            keys.push(key);
+          }
+
+          return new ArrayIterator(keys, 'value');
```
Victorystick: I tested @zloirock's enumerate implementation in Chome and Firefox, and it seems like the possible keys which the iterator can yield are determined when `iter.next()` is first called. Properties defined afterwards are not included, nor any deleted properties.

```javascript
var obj = { a: 1, b: 2 }, iter = Reflect.enumerate(obj);
iter.next(); // { value: "a", done: false }

obj.c = 3;
iter.next(); // { value: "b", done: false }
iter.next(); // { value: undefined, done: true }
```
```javascript
var obj = { a: 1, b: 2 }, iter = Reflect.enumerate(obj);

obj.c = 3;
iter.next(); // { value: "a", done: false }
iter.next(); // { value: "b", done: false }
iter.next(); // { value: "c", done: false }
iter.next(); // { value: undefined, done: true }

```
```javascript
var obj = { a: 1, b: 2 }, iter = Reflect.enumerate(obj);

iter.next(); // { value: "a", done: false }
delete obj.b;
iter.next(); // { value: undefined, done: true }

```

ljharb: I think their implementation is wrong - https://people.mozilla.org/~jorendorff/es6-draft.html#sec-ordinary-object-internal-methods-and-internal-slots-enumerate clearly says if a property is deleted before it's iterated over, it's ignored, and that "if new properties are added to the object being enumerated during enumeration, the newly added properties are not guaranteed to be processed in the active enumeration." - so deleted properties must be ignored, but newly added ones don't necessarily.

This suggests that we might need, on each `next` call, to iterate via `for..in` or `Object.keys` until the "last seen key" is seen, and then return the key after that - ie, each `next` call returns something like `var keys = Object.keys(obj); var i = keys.indexOf(lastKey); lastKey = i; return keys[i + 1];`

zloirock: It was just an example to show that the hole should be ignored, and not the full implementation :)

Victorystick:
Reflect is getting more or less done. Time for another look?

ljharb:
Sweet, will take another look today!

es6-shim.js
```diff
@@ -205,7 +205,15 @@
     return result;
   };
 
+  var bindContext = function (func, context) {
```
ljharb: `var bindContext = Function.call.bind(Function.apply)`? I'm not even sure we need a shortcut variable defined here.

es6-shim.js
```diff
@@ -205,7 +205,15 @@
...
+    return function () {
+      return func.apply(context, arguments);
+    };
+  };
+
   var ES = {
+    Call: bindContext(Function.prototype.call, Function.prototype.apply),
```
ljharb: Typically `ES` is reserved for spec-internal operations - is `Call` one of them?

es6-shim.js
```diff
@@ -720,6 +728,13 @@
     defineProperty(Array, 'from', ArrayShims.from, true);
   }
 
+  // Given an argument x, it will return an IteratorResult object,
+  // with value set to x and done to false.
+  // Given no arguments, it will return an iterator completion object.
+  var iterator_result = function (x) {
+    return { value: x, done: !arguments.length };
```
ljharb: Please use `arguments.length === 0` so as to not rely on number-boolean coercion.

es6-shim.js
```diff
@@ -756,6 +771,64 @@
...
+  };
+
+  defineProperties(ObjectIterator.prototype, {
+    next: function () {
+      var key, array = this.array;
+      if (!(this instanceof ObjectIterator)) {
+        throw new TypeError('Not an ObjectIterator');
+      }
+
+      if (typeof array !== 'undefined') {
```
ljharb: When will `typeof array` be `undefined` here?

es6-shim.js
```diff
@@ -756,6 +771,64 @@
...
+          this.i++;
+
+          // The candidate key isn't defined on object.
+          // Must have been deleted, or object[[Prototype]]
+          // has been modified.
+          if (!(key in this.object)) {
+            continue;
+          }
+
+          switch (this.kind) {
```
ljharb: Please use an if/else instead of a switch - I have a strong bias against switch statements in JS (that we don't need to get into here)

es6-shim.js
```diff
@@ -756,6 +771,64 @@
...
+
+  defineProperties(ObjectIterator.prototype, {
+    next: function () {
+      var key, array = this.array;
+      if (!(this instanceof ObjectIterator)) {
+        throw new TypeError('Not an ObjectIterator');
+      }
+
+      if (typeof array !== 'undefined') {
+        // Keys not generated
```
ljharb: Could we put the "key generation" logic in a separate, closured function? Something like `this.array = getAllKeys(this.object);`

es6-shim.js
rwaldron: A good rule: only use switch when a program _must_ take advantage of multiple case fall-through, otherwise use if/else (as in this example).

es6-shim.js
```diff
@@ -756,6 +771,64 @@
...
+          case 'entry':
+            return iterator_result([key, this.object[key]]);
+          }
+        }
+
+        // End of key array.
+        this.array = void 0;
+      }
+
+      return iterator_result();
```
ljharb: I'm not sure I see the value of this here - why not just `return { done: true };`?

es6-shim.js
```diff
@@ -2097,6 +2170,257 @@
...
+      var Reflect = {};
+
+      var wrapObjectFunction = function (func) {
+        return function (target) {
+          if (!ES.TypeIsObject(target)) {
+            throw new TypeError('target must be an object');
+          }
+
+          try {
+            func.apply(Object, arguments);
```
ljharb: Wrapping everything in try/catch makes things really slow - is this necessary for every case?

es6-shim.js
```diff
@@ -2097,6 +2170,257 @@
...
+
+          return internal_get(target, key, receiver);
+        },
+
+        getOwnPropertyDescriptor: throwUnlessTargetIsObject(Object.getOwnPropertyDescriptor),
+
+        getPrototypeOf: throwUnlessTargetIsObject(Object.getPrototypeOf),
+
+        has: function has(target, key) {
+          if (!ES.TypeIsObject(target)) {
```
ljharb: Could we wrap `function has` in `throwUnlessTargetIsObject` instead, so it's consistent?

es6-shim.js
```diff
@@ -2097,6 +2170,257 @@
...
+        enumerate: function enumerate(target) {
+          if (!ES.TypeIsObject(target)) {
+            throw new TypeError('target must be an object');
+          }
+
+          return new ObjectIterator(target, 'key');
+        },
+
+        // Syntax in a functional form.
+        get: function get(target, key) {
```
ljharb: Could we wrap `function get` in `throwUnlessTargetIsObject` instead, so it's consistent?

es6-shim.js
```diff
@@ -2097,6 +2170,257 @@
...
+          if (desc && !desc.configurable) {
+            return false;
+          }
+
+          // Will return true.
+          return delete target[key];
+        },
+
+        enumerate: function enumerate(target) {
+          if (!ES.TypeIsObject(target)) {
```
ljharb: Could we wrap `function enumerate` in `throwUnlessTargetIsObject` instead, so it's consistent?

es6-shim.js
```diff
@@ -2097,6 +2170,257 @@
...
+        },
+
+        defineProperty: wrapObjectFunction(Object.defineProperty),
+
+        // When deleting a non-existant or configurable property,
+        // true is returned.
+        // When attempting to delete a non-configurable property,
+        // it will return false.
+        deleteProperty: function deleteProperty(target, key) {
+          if (!ES.TypeIsObject(target)) {
```
ljharb: Could we wrap `function deleteProperty` in `throwUnlessTargetIsObject` instead, so it's consistent?

es6-shim.js
```diff
@@ -2097,6 +2170,257 @@
...
+            while (p) {
+              if (object === p) {
+                return false;
+              }
+
+              p = Reflect.getPrototypeOf(p);
+            }
+          }
+
+          Object.setPrototypeOf(object, proto);
```
ljharb: What happens in a browser where `Object.setPrototypeOf` is not defined (because it sometimes can't be shimmed)? I'd think `Reflect.setPrototypeOf` should similarly not be shimmed in the first place, in this case.

es6-shim.js
```diff
@@ -2097,6 +2170,257 @@
...
+          if (proto === Reflect.getPrototypeOf(object)) {
+            return true;
+          }
+
+          // Cannot alter prototype if object not extensible.
+          if (!Reflect.isExtensible(object)) {
+            return false;
+          }
+
+          // Ensure that we do not create a circular prototype chain.
```
ljharb: I wonder if this would also be more readable as a separate closured function?

test/reflect.js
```diff
@@ -0,0 +1,483 @@
...
+
+var exported = require('../');
+
+describe('Reflect', function () {
+  var object = {
+    something: 1,
+    _value: 0
+  };
+
+  Object.defineProperties(object, {
```
ljharb: These tests need to run in ES3, where `Object.defineProperties` won't be available. Please ensure that any tests that rely on property descriptors being supported are skipped in browsers that don't support them.

test/reflect.js
```diff
@@ -0,0 +1,483 @@
...
+    },
+
+    bool: {
+      value: true
+    }
+  });
+
+  var testCallableThrow = function (func) {
+    [null, undefined, 1, 'string', true, [], {}].forEach(function (item) {
+      expect(func.bind(null, item)).to.throw(TypeError);
```
ljharb: `function () { func(item); }` would be a bit faster here, and wouldn't look significantly messier imo. Just a thought.

Either way, `func.bind(null, item)` or `function () { func(item); }` should be defined outside of the loop, to avoid creating a function in each iteration. Same below.

test/reflect.js
```diff
@@ -0,0 +1,483 @@
...
+    });
+
+    it('works also with redefined apply', function () {
+      expect(Reflect.apply(Array.prototype.push, [1, 2], [3, 4, 5])).to.equal(5);
+
+      function F(a, b, c) {
+        return a + b + c;
+      }
+
+      F.apply = false;
```
ljharb: We should probably redefine it to be a function, so that any `function` checks would return `true`, making the test more robust.

test/reflect.js
```diff
@@ -0,0 +1,483 @@
...
+
+    it('throws if target isn\'t callable', function () {
+      testCallableThrow(function (item) {
+        return Reflect.apply(item, null, []);
+      });
+    });
+
+    it('works also with redefined apply', function () {
+      function C(a, b, c) {
+        this.qux = a + b + c;
```
ljharb: Not a big deal at all, but `[a, b, c].join('|')` might be a bit clearer since there'd be a separator.

test/reflect.js
```diff
@@ -0,0 +1,483 @@
...
+    it('is a function', function () {
+      expect(typeof Reflect.isExtensible).to.equal('function');
+    });
+
+    it('returns true for plain objects', function () {
+      expect(Reflect.isExtensible({})).to.equal(true);
+      expect(Reflect.isExtensible(Object.preventExtensions({}))).to.equal(false);
+    });
+
+    it('throws if the target isn\'t an object', function () {
```
ljharb: feel free to use `’` here to both avoid the need to escape, and to please the typographically inclined

test/reflect.js
```diff
@@ -0,0 +1,483 @@
...
+
+      obj[1] = 'first';
+
+      var result = Object.getOwnPropertyNames(obj);
+
+      expect(Reflect.ownKeys(obj)).to.deep.equal(result);
+      expect(result).to.deep.equal(['1', 'foo', 'bar']);
+    });
+
+    (typeof Symbol === 'function' ? it : xit)('symbols come last', function () {
```
ljharb: Let's make a var at the top of the file, something like `var ifSymbolsIt = typeof Symbol === 'function' ? it : xit` so that multiple tests can cleanly do `ifSymbolsIt('does something', …)`

test/reflect.js
```diff
@@ -0,0 +1,483 @@
...
+      expect(Reflect.setPrototypeOf(obj, Array.prototype)).to.equal(true);
+      expect(obj).to.be.an.instanceOf(Array);
+
+      expect(obj.toString).not.to.equal(undefined);
+      expect(Reflect.setPrototypeOf(obj, null)).to.equal(true);
+      expect(obj.toString).to.equal(undefined);
+    });
+
+    it('is returns false on failure', function () {
+      var obj = Object.freeze({});
```
ljharb: This test needs to be skipped in environments where `Object.freeze` is unavailable.

ljharb:
@Victorystick OK, I think all the things I commented on are minor, the biggest being that both the `es6-shim` and its tests need to run all the way down to ES3 browsers, so everything that's not shimmable in ES3 or ES5 needs to be feature-detected at runtime. Almost there :-)

Thanks, this looks awesome! I'm excited to merge this (after the changes above, and after I do some cross-browser testing)

es6-shim.js
Victorystick: I believe so. "The [[[Call]] internal method](https://people.mozilla.org/~jorendorff/es6-draft.html#sec-built-in-function-objects-call-thisargument-argumentslist) for a built-in function object F is called with parameters thisArgument and argumentsList."

es6-shim.js
Victorystick: Sure. I just mimicked the ArrayIterator. State enum?

es6-shim.js
Victorystick: Right. I though it was a neat way of avoiding to alias this.kind to kind, as well as being easier to read. But, I'll change it.

es6-shim.js
Victorystick: All the other iterators return `{ value: void 0, done: true }`, for spec compliance?
Using a function like so should have the same result, as well as minifying better. Although I understand that is not a primary goal.

es6-shim.js
Victorystick: The Reflect methods should return a boolean value depending on the success/failure of an internal operation. The regular Object methods throw an error if the internal operation returns false, otherwise it returns the target object.

The advantage of this approach is that it is much shorter and simpler, but should behave the same.

es6-shim.js
ljharb: Fair point - I can get behind making sure they're all consistent. Maybe we could make a separate function `iteratorDoneResult` that returns `{ value: void 0, done: true }`, and then make sure to use that throughout the shim?

es6-shim.js
Victorystick: Since the definition of Reflect is contained in the same `if (supportsDescriptors) {`-block as the `Object.setPrototypeOf` shim, I assumed setPrototypeOf to exist.

test/reflect.js
Victorystick: The point is to have a function with a non-function apply property. If we don't guard against it, it would be possible to prevent Reflect.apply from functioning if it is implemented simply as `func.apply(context, args)`. Right?

Victorystick:
Have you also noticed that the Travis CI builds often fail because `sh: 1: jscs: not found`?

es6-shim.js
Victorystick: If you think it is clearer with a specialized one for end of iteration. I considered making another PR after this one, with all iterators using some kind of iteratorResult function.

es6-shim.js
ljharb: Good call - https://people.mozilla.org/~jorendorff/es6-draft.html#sec-call-f-v-args is in section 7, so this is good.

es6-shim.js
ljharb: Sounds good, let's defer that til after.

ljharb:
Yes :-/ that's an issue with `jscs`'s dependency tree. For now I just manually restart those builds and they pass eventually.

test/reflect.js
ljharb: Sure - but the same hazard exists if it *is* a function property that doesn't have the same semantic as "apply". Perhaps two tests, one with it as a non-function, and one with it as a noop?

es6-shim.js
ljharb: I wonder if it would be better, instead of keeping an array of all keys and iterating over it, to continually `unshift` it in a loop? That way, when it's not an array, you need to populate it, and when it's an empty array, you know it's done.

That would eliminate the need for a state enum as well as 1 of the 3 states, and it would also eliminate the need for `this.i`.

es6-shim.js
Victorystick: I think that `ES.Call` could be really useful, and if it can be written as `Function.call.bind(Function.apply)`, I think `bindContext` might as well be removed.

ljharb: Sounds good to me!

ljharb: I've released https://www.npmjs.com/package/es-abstract for the `es7-shim`, which I eventually intend to add to the `es6-shim` once I get a build process going. For now, please consider any ES function inside `es-abstract` fair game for copy-paste into the `es6-shim`, including `ES.Call` :-) Thanks for this thread, it helped lead me to the idea.

ljharb:
I think I'm primarily waiting for `bindContext` to be removed before a final rereview and merge.

es6-shim.js
```diff
@@ -2097,6 +2170,233 @@
...
+        }),
+
+        // Syntax in a functional form.
+        get: throwUnlessTargetIsObject(function get(target, key) {
+          var receiver = arguments.length > 2 ? arguments[2] : target;
+
+          return internal_get(target, key, receiver);
+        }),
+
+        getOwnPropertyDescriptor: throwUnlessTargetIsObject(Object.getOwnPropertyDescriptor),
```
ljharb: The es6-shim needs to work in ES3 engines, which don't have property descriptors, a shimmable `setPrototypeOf`, `Object.isExtensible`, and a number of others.

Each Reflect method that can't work as-is in shimmed ES3 needs to be conditionally added.

es6-shim.js
```diff
@@ -2097,6 +2170,233 @@
...
+        // Basically the result of calling the internal [[OwnPropertyKeys]].
+        // Concatenating propertyNames and propertySymbols should do the trick.
+        // This should continue to work together with a Symbol shim
+        // which overrides Object.getOwnPropertyNames and implements
+        // Object.getOwnPropertySymbols.
+        ownKeys: throwUnlessTargetIsObject(function ownKeys(target) {
+          var keys = Object.getOwnPropertyNames(target);
+
+          if (ES.IsCallable(Object.getOwnPropertySymbols)) {
+            keys = keys.concat(Object.getOwnPropertySymbols(target));
```
ljharb: Would this be faster as `keys.push.apply(keys, Object.getOwnPropertySymbols(target))`?

ljharb:
OK, we're down to the primary blocker being that any `Reflect` method that can't work in ES3 (ie, ones that use property descriptors, or setPrototypeOf, etc) need to be conditionally added.

Thanks a lot for being persistent on this, this is going to be great to have!

zloirock:
Latest spec draft adds `newTarget` argument to `Reflect.construct`. It can't be fully polyfilled, but result should be created from `newTarget.prototype`.

ljharb:
Anything that can't be fully polyfilled in any engine needs to go in the es6-sham, not the es6-shim. However, if it can be fully polyfilled in true ES5 engines (via `Object.create`) then that's great

es6-shim.js
Victorystick: The definition of Reflect is inside the `if (supportsDescriptors)` block along with Set and Map. Therefore I suppose that the descriptor support exists. Can't I be sure Object.getOwnPropertyDescriptor exists?

es6-shim.js
Victorystick: We could very well change this.

Victorystick:
I set out with the intention of polyfilling Reflect in an ES5 environment, since it is so dependent on the Object property methods.

es6-shim.js
ljharb: Ah, I didn't realize. In that case, then the comment changes to be, any `Reflect` methods that *don't* require descriptors should be defined even when they're not supported :-) but that can be a followup, since that at least won't be broken anywhere. Thanks for clarifying!

ljharb:
OK - so this looks good to go, as long as it's freshly rebased on top of the latest master! \( ﾟ◡ﾟ)/

Followup item is to pull as much `Reflect` functionality as possible into ES3 engines.

Victorystick:
This should be it. :+1:

ljharb:
@Victorystick It looks great and I'm ready to merge it, but I still see a merge commit (37e8818db3b4e03a9b341106381db9a1cc38b012) - can you do a *rebase* onto latest master (as opposed to a merge with latest master)?

ljharb:
I'm anxious to merge this, so unless you're already in progress I'll make a new branch and rebase it myself (your name will remain on all the commits).

ljharb:
Merged via dd6b1b254fb6e08045a9f0bdce3620457581528f

ljharb:
Thanks so much for contributing this, and working through all the back and forths with me!

Victorystick:
No worries. Thank you!

PR #323, opened 2015-02-26
Math.log2 should be exact for exact powers

Yaffle:
Math.log2(1.0 * (1 << 29)) === 29.000000000000004 in Firefox

### DIFF
--- a/es6-shim.js
+++ b/es6-shim.js
@@ -1499,7 +1499,7 @@
 
     log2: function (value) {
       if (value === 0) {
-        return -1 / 0;
+        return -Infinity;
       }
       var e = Math.min(Math.floor(Math.log(value) / Math.LN2), 1023);
       return Math.log(value / Math.pow(2, e)) * Math.LOG2E + e;
--- a/es6-shim.js
+++ b/es6-shim.js
@@ -1501,7 +1501,10 @@
       if (value === 0) {
         return -Infinity;
       }
-      var e = Math.min(Math.floor(Math.log(value) / Math.LN2), 1023);
+      if (value === Infinity) {
+        return Infinity;
+      }
+      var e = Math.trunc(Math.log(value) * Math.LOG2E - Math.sign(value - 1) * 0.5);
       return Math.log(value / Math.pow(2, e)) * Math.LOG2E + e;
     },
 
--- a/es6-shim.js
+++ b/es6-shim.js
@@ -1498,7 +1498,17 @@
     },
 
     log2: function (value) {
-      return Math.log(value) * Math.LOG2E;
+      value = Number(value);
+      var y = Math.log(value) * Math.LOG2E;
+      var n = y | 0;
+      if (y - n > 0.5) {
+        n += 1;
+      }
+      if (y - n < -0.5) {
+        n -= 1;
+      }
+      var p = Math.pow(2, n);
+      return p <= value && y < n || p >= value && y > n ? n : y;
     },
 
     log10: function (value) {
@@ -1592,6 +1602,10 @@
   if (Math.imul(0xffffffff, 5) !== -5) {
     // Safari 6.1, at least, reports "0" for this value
     Math.imul = MathShims.imul;
+  }
+
+  if (Math.log2(536870912.0) !== 29) {
+    Math.log2 = MathShims.log2;
   }
 
   // Promises
--- a/es6-shim.js
+++ b/es6-shim.js
@@ -1504,8 +1504,8 @@
       if (value === Infinity) {
         return Infinity;
       }
-      var e = Math.trunc(Math.log(value) * Math.LOG2E - Math.sign(value - 1) * 0.5);
-      return Math.log(value / Math.pow(2, e)) * Math.LOG2E + e;
+      var e = Math.trunc(Math.log(value) / Math.LN2 - Math.sign(value - 1) * 0.5);
+      return Math.log(value / Math.pow(2, e)) / Math.LN2 + e;
     },
 
     log10: function (value) {
--- a/es6-shim.js
+++ b/es6-shim.js
@@ -1498,17 +1498,11 @@
     },
 
     log2: function (value) {
-      value = Number(value);
-      var y = Math.log(value) * Math.LOG2E;
-      var n = y | 0;
-      if (y - n > 0.5) {
-        n += 1;
+      if (value === 0) {
+        return -1 / 0;
       }
-      if (y - n < -0.5) {
-        n -= 1;
-      }
-      var p = Math.pow(2, n);
-      return p <= value && y < n || p >= value && y > n ? n : y;
+      var e = Math.min(Math.floor(Math.log(x) / Math.LN2), 1023);
+      return Math.log(x / Math.pow(2, e)) * Math.LOG2E + e;
     },
 
     log10: function (value) {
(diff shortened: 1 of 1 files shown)


cscott:
Really?  Does the spec say this should be exact?

es6-shim.js
```diff
@@ -1594,6 +1604,10 @@
     Math.imul = MathShims.imul;
   }
 
+  if (Math.log2(536870912.0) !== 29) {
+    Math.log2 = MathShims.log2;
```
ljharb: This would need to use `defineProperty` to preserve non-enumerability.

ljharb:
I'm all for making Math shims more accurate, even if the spec doesn't require it.

However, we'd need the following:
 - tests that ensure the behavior is correct
 - example engines that implement this correctly
 - a link to the relevant spec section that, at the least, leaves it to "implementation-dependent"

cscott:
Well, at least two projects that I know of have recently switched to `core-js` for their shims because `es6-shim` was "too slow".  Adding extra cruft to methods to obtain accuracy not required by the spec will not help `es6-shim` if it slows down the library for all users.

es6-shim.js
Yaffle: @ljharb , assignment preservers enumerability (previously I thought it sets all flags to "true", but it preservers it, seems)

Yaffle:
@ljharb , I have updated my patch a little

es6-shim.js
```diff
@@ -1498,7 +1498,11 @@
     },
 
     log2: function (value) {
-      return Math.log(value) * Math.LOG2E;
+      if (value === 0) {
+        return -1 / 0;
+      }
+      var e = Math.min(Math.floor(Math.log(value) / Math.LN2), 1023);
```
Yaffle: instead of 1023 there should be `Math.floor(Math.log(Number.MAX_VALUE) / Math.log(2))` constant

ljharb:
@cscott What is "too slow"? Which methods? Did they use profiling to determine that?

es6-shim.js
```diff
@@ -1498,7 +1498,11 @@
     },
 
     log2: function (value) {
-      return Math.log(value) * Math.LOG2E;
+      if (value === 0) {
+        return -1 / 0;
```
ljharb: Can this just be `-Infinity`? It's a keyword, so there's no advantage to the division by zero.

es6-shim.js
ljharb: if that's where `1023` comes from, then absolutely a named constant would be ideal.

es6-shim.js
ljharb: huh, indeed it does, disregard my comment then :-)

es6-shim.js
Yaffle: if you want

zloirock:
@ljharb 
> What is "too slow"? Which methods? Did they use profiling to determine that?

Simple example, in all modern environment fully complying with the specification `Map` and `Set`, `es6-shim` replaces to slow shim with O(n) search complexity for objects. Try with `es6-shim` and without:
```js
var array = [];
for(var i = 0; i < 100000; i++)array.push([{}, {}]);
array = array.concat(array);
console.time('Map test');
var map = new Map(array);
console.log('Map size: ' + map.size);
console.timeEnd('Map test');
```

ljharb:
@zloirock they're not actually fully complying if the shim is replacing them (zero engines fully comply at the moment). But yes, that is a limitation, since it's impossible to a) not mutate the objects and also b) have O(1) lookup for them.

However, I've filed https://github.com/paulmillr/es6-shim/issues/326 so that in this situation, we use the native `Map` for O(1) lookup, which should alleviate the speed issue.

ljharb:
@Yaffle This code is totally fine now, thanks! Could you also add a test that fails without your changes?

zloirock:
@ljharb
> they're not actually fully complying if the shim is replacing them

they're  fully complying, but your subclassing detect is broken :)

> that is a limitation, since it's impossible

but why `core-js` and some other shims polyfill it? :D

ljharb:
@zloirock `core-js` must be modifying the objects. I'd be happy to accept a pull request if you know of a way to do it, or a link to their implementation. Nothing must ever mutate unless the spec requires it to, so if it's mutating, it's violating the spec.

As for the subclassing detection, please file an issue or open a PR, but I don't think it's broken - you can't actually subclass the native `Map`/`Set` implementations at the moment precisely because of how the test fails without the shim ¯\_(ツ)_/¯

Yaffle:
@zloirock  object mutation is not good and may even affect performance (I hope yo read this page - https://developers.google.com/v8/design )
Anyway, this depends on the use case: if some user wants to store objects in Map AND thinks that it is OK to mutate them, then He can use `core-js`.

zloirock:
@Yaffle broken and slow collections much worse :)

ljharb:
Let's debate this on another issue, this isn't the one for it, but no, slow is infinitely better than broken (ours is not broken in any way), and mutation is absolutely broken (and as pointed out, insanely deoptimizes objects in v8, so it'd end up being slower in many ways as well).

zloirock:
@ljharb [this line](https://github.com/paulmillr/es6-shim/blob/master/es6-shim.js#L42) should not work in ES6. All modern browsers, includes V8, contains native collections with correct logic. All from me :)

zloirock:
Oh, it's more: https://github.com/babel/babel/issues/1172

ljharb:
@zloirock How then can you subclass `Map`/`Set` without `class`? I'll confirm with TC39 but I'm reasonably sure that's, at best, an unintended side effect.

zloirock:
@ljharb see link. It's only emulation, but correct in most case and much better replacing correct collection to broken.

es6-shim.js
Yaffle: I need to think how to change this.
Seems, this is a little wrong way of integer part calculation...

Yaffle:
@ljharb , examples of engines: the engine of Edge, the engine of Chrome, the engine of Firefox (only for log10, https://bugzilla.mozilla.org/show_bug.cgi?id=1070517 ), the engine of Safari

```
// fdlibm

Math.getExponent = function (x) {
  var e = Math.floor(Math.log(x) / Math.log(2) + 0.5);
  return Math.pow(2, e) > x ? e - 1 : e;
};

Math.log10 = function (x) {
  if (x === 0) {
    return -1 / 0;
  }
  if (x === 1 / 0) {
    return 1 / 0;
  }
  var e = Math.getExponent(x);
  if (e < 0) {
    e += 1;
  }
  return Math.LOG10E * Math.log(x / Math.pow(2, e)) + e * 3.69423907715893078616e-13 + e * 3.01029995663611771306e-01;
};

Math.log2 = function (x) {
  if (x === 0) {
    return -1 / 0;
  }
  if (x === 1 / 0) {
    return 1 / 0;
  }
  var e = Math.getExponent(x);
  if (e < 0) {
    e += 1;
  }
  return Math.LOG2E * Math.log(x / Math.pow(2, e)) + e * 0 + e * 1;
};
```

Yaffle:
@ljharb , example engines that implement this correctly: the engine of Edge, Chrome, Safari, Firefox (with an exception of `Math.log2` results under Windows)

PR #325, opened 2015-04-02
fix for Math.round

Yaffle:


es6-shim.js
```diff
@@ -1585,13 +1585,13 @@
   var expm1OfTen = Math.expm1(10);
   defineProperty(Math, 'expm1', MathShims.expm1, expm1OfTen > 22025.465794806719 || expm1OfTen < 22025.4657948067165168);
 
-  var roundHandlesBoundaryConditions = Math.round(0.5 - Number.EPSILON / 4) === 0 && Math.round(-0.5 + Number.EPSILON / 3.99) === 1;
```
ljharb: I'd prefer to retain the original test, since that's what's been verified to fail on specific browser instances. Adding your tests is great though.

test/math.js
```diff
@@ -439,6 +439,14 @@ describe('Math', function () {
       expect(Object.is(-0, Math.round(-0.5 + Number.EPSILON / 3.99))).to.equal(true);
       expect(Object.is(-0, Math.round(0 - Number.EPSILON / 3.99))).to.equal(true);
     });
+
+    it('returns 1 / Number.EPSILON + 1 for 1 / Number.EPSILON + 1', function () {
+      expect(Math.round(1 / Number.EPSILON + 1)).to.equal(1 / Number.EPSILON + 1);
```
ljharb: Can you elaborate a bit more on what current browsers do for these inputs, and what (if anything) the spec requires for it?

es6-shim.js
Yaffle: original test is wrong `-0.5 + Number.EPSILON / 3.99` will have same value as `-0.5 + Number.EPSILON / 4`

test/math.js
Yaffle: Safari 8, IE 11, Opear 12 and FF < 12 (?) are buggy here, 

Spec says "Returns the Number value that is closest to x and is equal to a mathematical integer"
All values greater than `4503599627370496` are already integers. Browsers are buggy, because internally they do `Math.floor(x + 0.5)`, possibly, so the result is converted to the next integer sometimes.

Yaffle: `1 / Number.EPSILON + 1` - is the smallest number 4503599627370497, for which `Math.floor(x + 0.5)` gives wrong result (4503599627370497 + 0.5 === 4503599627370498)
and `2 / Number.EPSILON - 1` is the largest number for which `Math.flor(x + 0.5)` gives wrong result.

Yaffle: This all is true for `double-precision 64-bit binary format IEEE 754` with `round to nearest, ties to even` mode, which are used by ECMAScript (see https://people.mozilla.org/~jorendorff/es6-draft.html#sec-ecmascript-language-types-number-type)

es6-shim.js
ljharb: You mean conceptually? The reason I put both is because I found engines where in fact they aren't the same.

ljharb:
Great, thanks for the explanation! If you can restore the original feature tests in the main shim file, and add a new, separate variable that represents your new feature tests, I'd be happy to merge this!

ljharb:
@Yaffle why did you close this?

Yaffle:
@ljharb , too many commits... 
I need to a add a comment message about buggy browsers also

ljharb:
No need to clutter up the repo with a second PR, just rebase this one and force push :-)

Yaffle:
@ljharb , I do not know how to rebase repo.

ljharb:
@Yaffle While I suggest you research how and learn (it's an awesome and powerful tool), you can also `git branch -m patch-2 tmp ; git checkout master ; git checkout -b patch-2` and then check out any changes you want from `tmp`, and force push `patch-2` which will update the PR accordingly.

Yaffle:
@ljharb thanks for help with git

Yaffle:
If you are interested - I am trying to add tests to test262: https://github.com/tc39/test262/pull/219

es6-shim.js
```diff
@@ -1580,14 +1580,13 @@
...
+  var roundDoesNotIncreaseIntegers = Math.round(1 / Number.EPSILON + 1) === 1 / Number.EPSILON + 1;
   defineProperty(Math, 'round', function round(x) {
-    if (-0.5 <= x && x < 0.5 && x !== 0) {
-      return Math.sign(x * 0);
-    }
-    return origMathRound(x);
-  }, !roundHandlesBoundaryConditions);
+    var y = Math.floor(x);
+    return x - y < 0.5 ? y : (y === -1 ? -0 : y + 1);
+  }, !roundHandlesBoundaryConditions && !roundDoesNotIncreaseIntegers);
```
ljharb: Shouldn't this be an `||`, so that a failure of either test overwrites the implementation?

es6-shim.js
```diff
@@ -1580,14 +1580,13 @@
...
-  var origMathRound = Math.round;
+  var roundDoesNotIncreaseIntegers = Math.round(1 / Number.EPSILON + 1) === 1 / Number.EPSILON + 1;
   defineProperty(Math, 'round', function round(x) {
-    if (-0.5 <= x && x < 0.5 && x !== 0) {
-      return Math.sign(x * 0);
-    }
-    return origMathRound(x);
-  }, !roundHandlesBoundaryConditions);
+    var y = Math.floor(x);
+    return x - y < 0.5 ? y : (y === -1 ? -0 : y + 1);
```
ljharb: Nested ternaries are really hard to read. Could you spread this out onto multiple lines, using multiple variables, so it's a bit clearer what's going on? Also, adding parentheses would help (in `x - y < 0.5` for example).

test/math.js
```diff
@@ -439,6 +439,14 @@ describe('Math', function () {
       expect(Object.is(-0, Math.round(-0.5 + Number.EPSILON / 3.99))).to.equal(true);
       expect(Object.is(-0, Math.round(0 - Number.EPSILON / 3.99))).to.equal(true);
     });
+
```
ljharb: This trailing whitespace is what failed the style checks.

es6-shim.js
Yaffle: yes, you are right

test/math.js
Yaffle: hm... is it so important?

ljharb: Yes, that's why it's part of the tests.

es6-shim.js
```diff
@@ -1580,14 +1580,14 @@
   var expm1OfTen = Math.expm1(10);
   defineProperty(Math, 'expm1', MathShims.expm1, expm1OfTen > 22025.465794806719 || expm1OfTen < 22025.4657948067165168);
 
+  // Safari 8, Internet Explorer 11, Opear 12
```
ljharb: s/Opear/Opera but I can fix this later too

ljharb:
OK, if we could condense this down to 1 or 2 commits on top of latest master one last time, I'll merge it in ASAP :-) From your explanations, I can figure out how to name the constants later, so thanks for that!

Yaffle:
@ljharb 
>You mean conceptually? The reason I put both is because I found engines where in fact they aren't the same.
No, I am not.
The engine for which `0.5 - Number.EPSILON / 4` is not equal to 0.5 - Number.EPSILON / 3.99` (when Number.EPSILON is defined), probably, has much bigger problems with floating point arithmetic.

`0.5 - Number.EPSILON / 4 * 1` <= `0.5 - Number.EPSILON / 3.99` <= `0.5 - Number.EPSILON / 4 * (4 / 3.99)`, BUT there should exist only next numbers:
`0.5 - Number.EPSILON / 4 * 1`
`0.5 - Number.EPSILON / 4 * 2`
And so this value SHOULD be equal to `0.5 - Number.EPSILON / 4 * 1`.

Currently, you are replacing `Math.round` with your implementation even in Chrome and Firefox, where it is not buggy, seems.

Yaffle:
@ljharb ?

ljharb:
@Yaffle I would be happy to accept a PR with failing tests, or improvements - I'm not sure I grasp exactly why there's an issue, and since the spec does not mandate any precision, I'm afraid I won't have time to get to it quickly (absent a PR).

Yaffle:
@ljharb 1) you are replacing not buggy `Math.round` in Chrome; 2) you have strange code

PR #327, opened 2015-04-08
more efficient Math.clz32

Yaffle:
see http://jsperf.com/read-leading-zeros/23

### DIFF
--- a/es6-shim.js
+++ b/es6-shim.js
@@ -1438,7 +1438,7 @@
       if (number === 0) {
         return 32;
       }
-      return 31 - Math.floor(Math.log(number) * Math.LOG2E + (1 / 2 / 4294967296));
+      return 31 - Math.floor(Math.log(number + 0.5) * Math.LOG2E);
     },
 
     cosh: function (value) {
--- a/es6-shim.js
+++ b/es6-shim.js
@@ -1438,7 +1438,7 @@
       if (number === 0) {
         return 32;
       }
-      return 32 - 1 - Math.floor(Math.log(number) * Math.LOG2E + (1 / 4294967296));
+      return 31 - Math.floor(Math.log(number) * Math.LOG2E + (1 / 2 / 4294967296));
     },
 
     cosh: function (value) {
--- a/es6-shim.js
+++ b/es6-shim.js
@@ -1438,7 +1438,7 @@
       if (number === 0) {
         return 32;
       }
-      return 32 - (number).toString(2).length;
+      return 32 - 1 - Math.floor(Math.log(number) * Math.LOG2E + (1 / 4294967296));
     },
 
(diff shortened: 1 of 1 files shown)


es6-shim.js
```diff
@@ -1438,7 +1438,7 @@
       if (number === 0) {
         return 32;
       }
-      return 32 - (number).toString(2).length;
+      return 32 - 1 - Math.floor(Math.log(number) * Math.LOG2E + (1 / 4294967296));
```
ljharb: What does `4294967296` represent? Could `1 / 4294967296` be stored in a named constant outside the shim function?

Also, why not `31 - `?

Some comments explaining why this works would be helpful also.

Yaffle: `Math.floor(Math.log(value) * Math.LOG2E)` is computed with some `error` and we need to fix this error. If we add something bigger than error, than `floor` should give values not lesser, than true result. But this something should not too big, otherwise we will get value one bigger, than true result. 
Also, it seems to me, `log(integer + 1) - log(integer) >= log(MAX_INTEGER+ 1) - log(MAX_INTEGER) >= 1 / (MAX_INTEGER + 1)`.
So "something" = `1 / (MAX_INTEGER + 1) / 2` should be good.
Seems, `Math.log` should be very inaccurate (with relative error > 2000 ulp (?)) to make this implementation incorrect.

es6-shim.js
```diff
@@ -1438,7 +1438,7 @@
       if (number === 0) {
         return 32;
       }
-      return 32 - (number).toString(2).length;
+      return 31 - Math.floor(Math.log(number) * Math.LOG2E + (1 / 2 / 4294967296));
```
ljharb: Can `1 / 2 / 4294967296` be cached in a variable outside the shimmed function, named so as to describe what it represents?

Yaffle: I think, it is not needed to put it outside. And it is hard for me to think of name...

ljharb: That's why I'm asking - if it's hard for you, it'll be infinitely harder for me :-) and the name will help with future understanding.

Yaffle: OK, seems, everything can be simpler
What about new code?

Yaffle:
OK, seems, everything can be simpler
What about new code?

ljharb:
The code looks great, and is clearly much faster than the existing implementation.

However, according to the jsperf, the overall fastest is "usingBitwiseOperations", not "usingLogarithm". Any reason not to use the bitwise method?

Yaffle:
@ljharb , bitwise method is bigger and uses 1) bitwise operations; 2) hardcoded precomputed constants

ljharb:
OK, let's merge this now, and consider the bitwise approach in the future.

Issue #330, opened 2015-04-10
Accuracy of Math.cbrt


Yaffle:
I have added 2 lines of code:

```
var cbrt = function (value) {
  value = Number(value);
  if (value === 0) { return value; }
  var negate = value < 0, result;
  if (negate) { value = -value; }
  result = Math.pow(value, 1 / 3);
  // from http://en.wikipedia.org/wiki/Cube_root#Numerical_methods
  result = (value / (result * result) + 2 * result) / 3;
  return negate ? -result : result;
};

var error = function (x, value) {
  return Math.abs(1 - x / value) / Number.EPSILON; // error in number of ulps (http://en.wikipedia.org/wiki/Unit_in_the_last_place)
};

console.log(error(Math.pow(2, 341), Math.cbrt(Math.pow(2, 341 * 3))));
console.log(error(Math.pow(2, 341), cbrt(Math.pow(2, 341 * 3))));

console.log(error(Math.pow(2, -340), Math.cbrt(Math.pow(2, -340 * 3))));
console.log(error(Math.pow(2, -340), cbrt(Math.pow(2, -340 * 3))));

//! WARNING: subnormals
console.log(error(Math.pow(2, -358), Math.cbrt(Math.pow(2, -358 * 3))));
console.log(error(Math.pow(2, -358), cbrt(Math.pow(2, -358 * 3))));

```

ljharb:
Closed via https://github.com/paulmillr/es6-shim/commit/5a7f3d770775a895b2962232622aeb7a49a8335e / #335

PR #335, opened 2015-04-14
a little more accurate Math.cbrt

Yaffle:
fix for #330

### DIFF
--- a/es6-shim.js
+++ b/es6-shim.js
@@ -1269,7 +1269,13 @@
       if (value === 0) { return value; }
       var negate = value < 0, result;
       if (negate) { value = -value; }
-      result = Math.pow(value, 1 / 3);
+      if (value === Infinity) {
+        result = Infinity;
+      } else {
+        result = Math.exp(Math.log(value) / 3);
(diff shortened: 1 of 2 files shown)


es6-shim.js
```diff
@@ -1269,7 +1269,13 @@
       if (value === 0) { return value; }
       var negate = value < 0, result;
       if (negate) { value = -value; }
-      result = Math.pow(value, 1 / 3);
+      if (value === Infinity) {
```
ljharb: This looks like the code path could be hit with `-Infinity` too, is that one accurate for the `exp` + `log` path?

Yaffle: -Infinity is converted to Infinity 2 lines above, so the "true" branch will be involved, and the "sign" will be set before return

ljharb: ah ha, so it will, thanks

ljharb:
This looks awesome, and your new tests already pass in everything I briefly checked.

Can you rebase this on top of the latest master (rather than merge) with (it looks like from your merge commit) `git rebase paulmillr/master`, and I'll get this merged?

Yaffle:
@ljharb , ok, thanks for the command

PR #337, opened 2015-04-16
fix for Math.asinh, Math.acos, Math.atanh....

Yaffle:


es6-shim.js
```diff
@@ -1240,30 +1240,30 @@
 
   var MathShims = {
     acosh: function acosh(value) {
-      var x = Number(value);
-      if (Number.isNaN(x) || value < 1) { return NaN; }
-      if (x === 1) { return 0; }
-      if (x === Infinity) { return x; }
-      return Math.log(x / Math.E + Math.sqrt(x + 1) * Math.sqrt(x - 1) / Math.E) + 1;
+      value = Number(value);
```
ljharb: Please don't revert the first line - reassigning values to arguments deoptimizes some JS engines.

es6-shim.js
```diff
@@ -1440,10 +1428,15 @@
   defineProperty(Math, 'log1p', MathShims.log1p, Math.log1p(-1e-17) !== -1e-17);
   // IE 11 TP has an imprecise asinh: reports Math.asinh(-1e7) as not exactly equal to -Math.asinh(1e7)
   defineProperty(Math, 'asinh', MathShims.asinh, Math.asinh(-1e7) !== -Math.asinh(1e7));
+  // IE 11 TP has an imprecise asinh: reports Math.asinh(-1e7) as not exactly equal to -Math.asinh(1e7)
```
ljharb: this message seems incorrect (and duplicated from line 1429)

ljharb:
Please add tests that would fail without all of these changes.

es6-shim.js
Yaffle: which one?
actually, I am OK with `var x = Number(value)`, but in some places you have it, in some not

ljharb: I just haven't yet gone through and made it consistent :-) `var x = Number(value)` seems to be a good standard for all the Math and Number functions.

es6-shim.js
```diff
@@ -1436,21 +1423,33 @@
     }
   };
   defineProperties(Math, MathShims);
-  // IE 11 TP has an imprecise log1p: reports Math.log1p(-1e-17) as 0
-  defineProperty(Math, 'log1p', MathShims.log1p, Math.log1p(-1e-17) !== -1e-17);
+  
+  var isNear = function (a, b) {
+    return Math.abs(1 - a / b) / Math.pow(2, -52) < 8;
```
Yaffle: I put `Math.pow(2, -52)` specially, because our "expected" values are computed only for binary64, so, possibly, it is better to not use Number.EPSILON here.

es6-shim.js
```diff
@@ -1436,21 +1423,33 @@
...
-  defineProperty(Math, 'expm1', MathShims.expm1, expm1OfTen > 22025.465794806719 || expm1OfTen < 22025.4657948067165168);
+  defineProperty(Math, 'expm1', MathShims.expm1, !isNear(MathShims.expm1(10), 22025.465794806718));
+  // IE 11 TP has an imprecise log1p: reports Math.log1p(-1e-17) as 0
+  defineProperty(Math, 'log1p', MathShims.log1p, !isNear(Math.log1p(-1e-17), -1e-17));
+  // old es6-shim
+  defineProperty(Math, 'sinh', MathShims.sinh, !isNear(Math.sinh(710), 1.1169973830808555e+308));
+  // node 0.11 has an imprecise Math.sinh with very small numbers
+  defineProperty(Math, 'sinh', MathShims.sinh, !isNear(Math.sinh(-2e-17), -2e-17));
+  // Chrome 40 on Windows has an imprecise Math.tanh with very small numbers
+  defineProperty(Math, 'tanh', MathShims.tanh, !isNear(Math.tanh(-2e-17), -2e-17));
```
Yaffle: ordered alphabetically

Yaffle:
@ljharb , for `test/math.js` it is possible to use code between line 1427 and line 1459, 
`almostEqual` at https://github.com/paulmillr/es6-shim/blob/master/test/math.js#L4 should be changed to use same code as `isNear`, so it will measure relative error, not absolute, and it will work better for small values

Yaffle:
:'(
I do not know how to work with `git`.

ljharb:
I've done the `var x` changes in master, so you'll want to rebase onto the latest master (and also reduce the 5 commits down).

As for `isNear` - I'd be happy to accept a separate PR to change the `almostEqual` expect implementation to be more precise, and update the associated tests.

Yaffle:
@ljharb 
continuation in https://github.com/paulmillr/es6-shim/pull/338

ljharb:
@Yaffle Please don't do this in the future - creating extra pull requests clutters up the repo, and they can't ever be deleted. Stick with one branch name per pull request.

PR #338, opened 2015-04-17
Math.acosh, Math.asinh, Math.atanh, Math.cosh, Math.sinh, Math.tanh

Yaffle:


test/math.js
```diff
@@ -1,10 +1,11 @@
...
 var Assertion = expect().constructor;
-Assertion.prototype.almostEqual = function (obj, precision) {
+var MATH_EXP_ERROR = Math.abs(1 - Math.exp(704.6589) / 1.0702171200481775e+306) / Number.EPSILON;
+Assertion.prototype.almostEqual = function (expected) {
   'use strict';
-  var allowedDiff = precision || 1e-11;
-  return this.within(obj - allowedDiff, obj + allowedDiff);
+  var ok = Math.abs(1 - this.__flags.object / expected) / Number.EPSILON < MATH_EXP_ERROR + 8;
+  this.assert(ok, 'expected #{this} to be almost equal ' + expected, 'expected #{this} to not be almost equal' + expected);
 };
```
Yaffle: this is not ideal, I do not know how to access private "this.__flags.object" and I did not test it well.

es6-shim.js
```diff
@@ -1370,32 +1364,29 @@
...
 
     sinh: function sinh(value) {
       var x = Number(value);
-      if (!globalIsFinite(x) || x === 0) { return x; }
-
       if (Math.abs(x) < 1) {
         return (Math.expm1(x) - Math.expm1(-x)) / 2;
       }
-      return (Math.exp(x - 1) - Math.exp(-x - 1)) * Math.E / 2;
+      return (Math.exp(x - 1) - Math.exp(-x - 1)) * (Math.E / 2);
```
Yaffle: brackets are important here - Math.sinh(710)

ljharb: I'm fine with adding these parentheses (brackets are `[ ]`), but mathematically it shouldn't make a difference.

Yaffle: mathematically - not, but from floating point perspective - should, because of overflow, also "(Math.E / 2)" in brackets allows js engines to precalculate constant value during, I think

ljharb: ah ha, good point :-)

test/math.js
ljharb: I think `this.flags('object')` might work better

Yaffle:
@ljharb , OK, what do you think now?

es6-shim.js
```diff
@@ -1416,41 +1410,55 @@
...
+      var sign = Math.sign(x);
+      var abs = Math.abs(x);
       if (abs < BINARY_32_MIN_VALUE) {
         return sign * roundTiesToEven(abs / BINARY_32_MIN_VALUE / BINARY_32_EPSILON) * BINARY_32_MIN_VALUE * BINARY_32_EPSILON;
       }
       // Veltkamp's splitting (?)
       var a = (1 + BINARY_32_EPSILON / Number.EPSILON) * abs;
       var result = a - (a - abs);
-      if (result > BINARY_32_MAX_VALUE || numberIsNaN(result)) {
+      if (result > BINARY_32_MAX_VALUE || Number.isNaN(result)) {
```
ljharb: Please revert this change - it's there intentionally, so that the ordering of shims doesn't matter.

es6-shim.js
```diff
@@ -1416,41 +1410,55 @@
...
       }
       return sign * result;
     }
   };
   defineProperties(Math, MathShims);
-  // IE 11 TP has an imprecise log1p: reports Math.log1p(-1e-17) as 0
-  defineProperty(Math, 'log1p', MathShims.log1p, Math.log1p(-1e-17) !== -1e-17);
+
+  // https://code.google.com/p/v8/issues/detail?id=3468
+  var MATH_EXP_ERROR = Math.abs(1 - Math.exp(704.6589) / 1.0702171200481775e+306) / Math.pow(2, -52);
```
ljharb: what do `704.6589` and `1.0702171200481775e+306` represent? can those be separate named variables?

es6-shim.js
```diff
@@ -1416,41 +1410,55 @@
...
+  var isAlmostEqual = function (a, b) {
+    return Math.abs(1 - a / b) / Math.pow(2, -52) < MATH_EXP_ERROR + 8;
```
ljharb: can `Math.pow(2, -52)` be stored in a separate variable so it's not recomputed every time?

es6-shim.js
```diff
@@ -1416,41 +1410,55 @@
...
+  // Chrome 40-43
+  defineProperty(Math, 'asinh', MathShims.asinh, !isAlmostEqual(Math.asinh(1.7976931348623157e+308), 710.475860073944));
+  // Chrome 40-43
+  defineProperty(Math, 'atanh', MathShims.atanh, !isAlmostEqual(Math.atanh(Math.pow(2, -1021)), Math.pow(2, -1021)));
   // Firefox 38 on Windows
-  defineProperty(Math, 'cbrt', MathShims.cbrt, Math.abs(1 - Math.cbrt(1e-300) / 1e-100) / Number.EPSILON > 8);
-  // node 0.11 has an imprecise Math.sinh with very small numbers
-  defineProperty(Math, 'sinh', MathShims.sinh, Math.sinh(-2e-17) !== -2e-17);
+  defineProperty(Math, 'cbrt', MathShims.cbrt, !isAlmostEqual(Math.cbrt(1e-300), 1e-100));
+  // old es6-shim
```
ljharb: what's this line here for?

ljharb:
The tests look good, and fail as expected on Chrome 42, node 0.8/0.10/0.11/0.12, and iojs 1.7.

es6-shim.js
Yaffle: So, `result !== result` is better, because it has no dependency on both `Number.isNaN` and `numberIsNaN`

es6-shim.js
Yaffle: @ljharb , I found the number `704.6589` experimentally, the relative error of `Math.exp` for this number is bigger than for any other number during my experiments.
Although, roughly the error grows with growth of the argument.
`1.0702171200481775e+306` - is an expected value for `Math.exp(7046589)`

es6-shim.js
Yaffle: well.... i think `Math.pow(2, -52)` is fast and can be even faster, `isAlmostEqual` is called after computation of other Math functions, computation of which is much more expensive

es6-shim.js
ljharb: For consistency, I'd prefer `numberIsNaN` so that the NaN checks can all be routed through the same code path.

es6-shim.js
ljharb: Separate variables with a comment explaining exactly those things would be wonderful!

es6-shim.js
```diff
@@ -1370,32 +1367,29 @@
...
-      if (x === Infinity) { return 1; }
-      if (x === -Infinity) { return -1; }
-      var a = Math.expm1(x);
-      var b = Math.expm1(-x);
-      if (a === Infinity) { return 1; }
-      if (b === Infinity) { return -1; }
-      return (a - b) / (Math.exp(x) + Math.exp(-x));
+      var a = Math.expm1(2 * x);
+      var b = Math.expm1(2 * x) + 2;
+      return a === b ? 1 : a / b;
```
ljharb: for what values is `a === b && a / b !== 1`?

Yaffle: Infinity

ljharb: oh snap again, good call. Is this faster enough that it's worth the far less clear version, than the previous which had explicit returns for ± `Infinity`?

Yaffle: no, 
`if (a === Infinity) { return 1; }` could be used

ljharb: I think keeping the explicit return checks for ± Infinity are a good idea - probably also the initial `NaN` and `0` checks too.

es6-shim.js
```diff
@@ -1241,29 +1241,29 @@
...
         return x;
       }
-      return x < 0 ? -Math.asinh(-x) : Math.log(x + Math.sqrt(x * x + 1));
+      if (x < 0) {
+        return -Math.asinh(-x);
+      }
+      if (x * x + 1 === x * x) {
+        return Math.log(x) + Math.log(2);
+      }
+      return Math.log1p(x + Math.expm1(Math.log1p(x * x) / 2));
```
lewisje: I was just about to post an issue recommending the use of a range-reduction method similar to that used for `Math.acosh` in the current shim:

```javascript
if (x < 0) { return -Math.asinh(-x); }
var sqx = x * x;
if (sqx <= inverseEpsilon - 1) { return Math.log(x + Math.sqrt(sqx + 1)); }
if (sqx <= Number.MAX_VALUE - 1) {
  var sqrtEpsilon = Math.sqrt(Number.EPSILON);
  return Math.log(x * sqrtEpsilon + Math.sqrt(square(x * sqrtEpsilon) + Number.EPSILON)) - Math.log(sqrtEpsilon);
}
var sqrtMax = Math.sqrt(Number.MAX_VALUE);
return Math.log(x / sqrtMax + Math.sqrt(square(x / sqrtMax) + 1 / Number.MAX_VALUE)) + Math.log(Number.MAX_VALUE) / 2;
```

However, this solution looks less hackish (also, I'm not sure whether the function call to `square()` would be worth it if the argument contains an operator, compared to running the inner operations twice).

Yaffle: @lewisje , 
Hi! Why is it better? Could you provide more details?
Your implementation is not accurate for small values (Math.pow(2, -1021)), 

What I do not like in my is : `if (x * x + 1 === x * x) {`. Possibly, I should use `x > Math.sqrt(2 / Number.EPSILON)` here.

lewisje: Sorry, by "this" I meant the line I was commenting on; that is, your idea is better than mine and I hope it is merged in.

test/math.js
Yaffle: @ljharb , I fixed by using `this.within`

es6-shim.js
Yaffle: OK, I followed this rule

es6-shim.js
Yaffle: @ljharb , it is hard to think about some name

es6-shim.js
Yaffle: removed

es6-shim.js
Yaffle: @ljharb , I cannot understand how explicit checks can help, I want to have number of "if" branches as small as possible. But I have replaced ternary operator with `if (a === Infinity && b === Infinity)` .

test/math.js
```diff
@@ -551,7 +565,11 @@ describe('Math', function () {
       expect(Math.sinh(-Infinity)).to.equal(-Infinity);
       expect(Math.sinh(-5)).to.almostEqual(-74.20321057778875);
       expect(Math.sinh(2)).to.almostEqual(3.6268604078470186);
-      expect(Math.sinh(-2e-17)).to.equal(-2e-17);
+      expect(Math.sinh(-2e-17)).to.almostEqual(-2e-17);
```
ljharb: In which browsers is `Math.sinh(-2e-17)` not precisely equal to `-2e-17`?

ljharb:
Thanks, this is looking great. I'll test it in a number of browsers before merging it.

test/math.js
Yaffle: @ljharb in a browser of 2150 year, which have Number.EPSILON < 2e-17 * 2e-17 / 3

es6-shim.js
```diff
@@ -1447,24 +1445,20 @@
...
-      if (Number.isNaN(x) || x === 0) { return x; }
-      if (x === Infinity) { return 1; }
-      if (x === -Infinity) { return -1; }
-      var a = Math.expm1(x);
-      var b = Math.expm1(-x);
-      if (a === Infinity) { return 1; }
-      if (b === Infinity) { return -1; }
-      return (a - b) / (Math.exp(x) + Math.exp(-x));
+      var a = Math.expm1(2 * x);
+      var b = Math.expm1(2 * x) + 2;
```
Yaffle: Why am I calculating `Math.expm1(2 * x)` twice?

ljharb: ha, good question

Yaffle:
any chance to finish this?

ljharb:
Absolutely - it'll need to be freshly rebased on top of master, and then I'll do a fresh round of browser testing - but I don't think it needs much more work.

Yaffle:
`Math.acosh`, `Math.asinh`, `Math.cosh`, `Math.sinh` are still not accurate ...
```
Math.acosh(1 + Number.EPSILON)
Math.asinh(Number.MAX_VALUE)
var EXP_NO_OVERFLOW = 710.4758600739439;
Math.cosh(EXP_NO_OVERFLOW + 1 / Math.LOG2E);
Math.sinh(EXP_NO_OVERFLOW + 1 / Math.LOG2E);
```

ljharb:
@Yaffle I'd love to bring this PR in if you want to freshly rebase it - it's been about 10 months :-)

Yaffle:
~2 years

ljharb:
@Yaffle yes, it's been 1 year since you closed the PR and deleted the branch without rebasing it, and since nobody else seems to understand this topic as well as you do, it will basically never be fixed.

I'd be delighted if you reopened the PR and rebased the branch, then it could get in very soon.

Yaffle:
I have put all implementations of not precise ES6 Math functions to the comment at
https://github.com/kangax/compat-table/issues/392#issuecomment-240196072

ljharb:
Yes, thank you, but a dump of implementations doesn't help me change anything. What is helpful is a set of test cases, justified with evidence (browser results, spec steps, etc), and then *also* an implementation if a good one isn't apparent.

Yaffle:
@ljharb,
1) a set of test cases:
`Math.asinh(1e+300)` should not be equal to +∞;
`Math.atanh(1e-300)` should not be equal to +0;
`Math.cosh(710)` should not be equal to +∞;
`Math.sinh(710)` should not be equal to +∞;
2) in good implementations (Chrome 54+, Firefox 48+, Safari 10+, Edge):
`Math.asinh(1e+300)`  returns ~`691.4686750787736`;
`Math.atanh(1e-300)` returns `1e-300`;
`Math.cosh(710)` returns `~1.1169973830808557e+308`;
`Math.sinh(710)` returns `~1.1169973830808557e+308`;
And so the relative error is very big.
Note: Chrome < 54 on Windows fails tests for Math.asinh and Math.atanh.
Chrome < 40 ? on Windows fails tests for Math.cosh and Math.sinh.
3) The implementations are in the my previous comment.

ljharb:
Thank you, that is much more helpful!

ljharb:
@Yaffle hmm, Chrome 38 and 30 both seem to pass the cosh tests; but do fail the sinh tests. I'll keep trying to find a cosh failure.

ljharb:
@Yaffle nvm, scratch that, i was on firefox :-)

ljharb:
@Yaffle thanks, those first 4 are taken care of in ec76d49; i missed the edit for the 5th and will follow up with a fix for that.

ljharb:
I'm unable to reproduce the `acosh` test failure on Chrome 53, 50, even down to 40.

Yaffle:
@ljharb , the test for `acosh` might look similar to test for `cbrt` at https://github.com/paulmillr/es6-shim/blob/ec76d49cb801a03fb0a98041e2bba190a41b8288/es6-shim.js#L2083 : 

`Math.abs(1 - (Math.acosh(1 + Number.EPSILON) / Math.sqrt(2 * Number.EPSILON))) / Number.EPSILON > 8`

ljharb:
thanks, i'll update that.

Yaffle:
@ljharb, thanks

ljharb:
k, should be all set.

Issue #239, opened 2014-04-01
Update Promise implementation to match new spec


cscott:
@domenic has updated the `Promise` spec: https://github.com/domenic/promises-unwrapping/issues/99#issuecomment-39188107

I do not believe there are any functional changes, so this isn't a critical update.  But I should update our implementation to better match the spec to make it more obviously correct (and to simplify, and to check @domenic's work!).

ljharb:
Fixed with 9542dda06c1812cefd14b5c6101a9f35ec1f2eda / #345

Issue #344, opened 2015-06-02
Update behavior of `Promise.resolve`.


cscott:
TC39 agreed to get rid of the hidden [[PromiseConstructor]] property from the spec and simplify Promise.resolve. The adopted proposal also makes `Promise.resolve` ignore "species":

https://esdiscuss.org/topic/fixing-promise-resolve

Opening an issue to track es6-shim adopting the new semantics.

ljharb:
It'd be ideal to also check on #239 as part of this work.

cscott:
Working on it right now (both this and #239, plus some incidental fixes to iterators).

ljharb:
@cscott there's another issue - see the tests in https://github.com/ljharb/es6-shim/commit/95f21e23872cee0d359d55159c6cc53166acc105 . It looks like our implementation (along with v8 and JSC, whereas Firefox and Bluebird pass) all fail these tests which are part of the spec. Mind taking a look at these too? :-)

cscott:
Those tests don't look right to me at all.  Can you point me at the
upstream source for them?

cscott:
In particular, the Promise Resolve function should execute the `then` in a
PromiseResolveThenableJob, which means it should *not* be synchronous.  Am
I misreading the spec?

ljharb:
You're correct - my initial commit was wrong, if you look at the latest one (comment has been updated) it tests that ".then" is called only once, as well as that the ".then" property is retrieved synchronously (es6-shim currently retrieves it asynchronously).

ljharb:
Fixed with 9542dda06c1812cefd14b5c6101a9f35ec1f2eda / #345

PR #345, opened 2015-06-04
Update Promise implementation to match latest ES6 spec.

cscott:
Includes the erratum for Promise.resolve discussed in #344.

Also fixes the Map/Set constructors, since the spec uses [[Construct]] now instead of a hidden create method, and the previous version misapplied `Symbol.species` in this role.

Also added ES.IteratorClose and updated a number of places where that is used in the latest spec, including in the implementation of `Array.from`.

Fixes issue #239 and #344.

### DIFF
--- a/es6-shim.js
+++ b/es6-shim.js
@@ -105,8 +105,9 @@
         Object.setPrototypeOf(o, Subclass.prototype);
         return o;
       };
+      Object.setPrototypeOf(Sub, C);
       Sub.prototype = create(C.prototype, {
-        constructor: { value: C }
+        constructor: { value: Sub }
       });
       return f(Sub);
     });
@@ -1925,7 +1926,7 @@
   // implementation is buggy in a number of ways.  Let's check subclassing
   // support to see if we have a buggy implementation.
   var promiseSupportsSubclassing = supportsSubclassing(globals.Promise, function (S) {
-    return S.resolve(42) instanceof S;
+    return S.resolve(42).then(function () {}) instanceof S;
   });
   var promiseIgnoresNonFunctionThenCallbacks = !throwsError(function () { globals.Promise.reject(42).then(null, 5).then(null, noop); });
   var promiseRequiresObjectContext = throwsError(function () { globals.Promise.call(3, noop); });
--- a/es6-shim.js
+++ b/es6-shim.js
@@ -149,6 +149,12 @@
 
   var Symbol = globals.Symbol || {};
   var symbolSpecies = Symbol.species || '@@species';
+  var defaultSpeciesGetter = function () { return this; };
+  var addDefaultSpecies = function (C) {
+    if (supportsDescriptors && !_hasOwnProperty(C, symbolSpecies)) {
+      Value.getter(C, symbolSpecies, defaultSpeciesGetter);
+    }
+  };
   var Type = {
     object: function (x) { return x !== null && typeof x === 'object'; },
     string: function (x) { return _toString(x) === '[object String]'; },
@@ -286,6 +292,11 @@
       return typeof x === 'function' && _toString(x) === '[object Function]';
     },
 
+    IsConstructor: function (x) {
+      // We can't tell callables from constructors in ES5
+      return ES.IsCallable(x);
+    },
+
     ToInt32: function (x) {
       return ES.ToNumber(x) >> 0;
     },
@@ -338,8 +349,9 @@
         // special case support for `arguments`
         return new ArrayIterator(o, 'value');
       }
-      var itFn = o[$iterator$];
+      var itFn = ES.GetMethod(o, $iterator$);
       if (!ES.IsCallable(itFn)) {
+        // Better diagnostics if itFn is null or undefined
         throw new TypeError('value is not an iterable');
       }
       var it = _call(itFn, o);
@@ -347,6 +359,43 @@
         throw new TypeError('bad iterator');
       }
       return it;
+    },
+
+    GetMethod: function (o, p) {
+      var func = ES.ToObject(o)[p];
+      if (func === void 0 || func === null) {
+        return void 0;
+      }
+      if (!ES.IsCallable(func)) {
+        throw new TypeError('Method not callable: ' + p);
+      }
+      return func;
+    },
+
+    IteratorComplete: function (iterResult) {
+      return !!(iterResult.done);
+    },
+
+    IteratorClose: function (iterator, completionIsThrow) {
+      var returnMethod = ES.GetMethod(iterator, 'return');
+      if (returnMethod === void 0) {
+        return;
+      }
+      var innerResult, innerException;
+      try {
+        innerResult = _call(returnMethod, iterator);
+      } catch (e) {
+        innerException = e;
+      }
+      if (completionIsThrow) {
+        return;
+      }
+      if (innerException) {
+        throw innerException;
+      }
+      if (!ES.TypeIsObject(innerResult)) {
+        throw new TypeError("Iterator's return method returned a non-object.");
+      }
     },
 
     IteratorNext: function (it) {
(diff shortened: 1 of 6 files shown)


es6-shim.js
```diff
@@ -285,6 +292,11 @@
       return typeof x === 'function' && _toString(x) === '[object Function]';
     },
 
+    IsConstructor: function (x) {
+      // We can't tell callables from constructors in ES5
+      return ES.IsCallable(x);
```
ljharb: we could do a "is not arrow function" check here too if we want, but that may not help much

es6-shim.js
```diff
@@ -337,8 +349,9 @@
         // special case support for `arguments`
         return new ArrayIterator(o, 'value');
       }
-      var itFn = o[$iterator$];
+      var itFn = ES.GetMethod(o, $iterator$);
       if (!ES.IsCallable(itFn)) {
```
ljharb: is this check needed now, since it's in `GetMethod`?

es6-shim.js
```diff
@@ -348,6 +361,43 @@
       return it;
     },
 
+    GetMethod: function (o, p) {
+      var func = (ES.ToObject(o))[p];
```
ljharb: why the wrapping parens? `ES.ToObject(o)[p]`, or even better, an intermediate variable for `ES.ToObject(o)`?

es6-shim.js
```diff
@@ -387,20 +467,32 @@
...
+      throw new TypeError('Constructor requires `new`: ' + defaultNewTarget.name);
     }
-    return object;
+    var proto = defaultNewTarget.prototype;
+    if (!ES.TypeIsObject(proto)) {
+      proto = defaultProto;
+    }
+    o = create(proto);
+    for (var name in slots) {
+      if (_hasOwnProperty(slots, name)) {
```
ljharb: Why not `Object.keys(slots).forEach` here? That will paper over a lot of browser bugs with `for..in` iteration.

es6-shim.js
```diff
@@ -593,53 +685,58 @@
   }
 
   var ArrayShims = {
-    from: function from(iterable) {
+    from: function from(items) {
```
ljharb: I wonder if it would be easier to get the `Array.from` and relevant internal changes in first, in a separate PR, and then get the `Promise` changes in afterwards?

test/promise/all.js
```diff
@@ -168,8 +169,10 @@ describe('Promise.all', function () {
           }, reject);
         });
       } else {
-        Promise.call(this, resolver);
+        self = new Promise(resolver);
       }
+      Object.setPrototypeOf(self, P.prototype);
```
ljharb: Please make sure this test bails out early if `Object.setPrototypeOf` is not shimmed in the current environment.

test/promise/evil-promises.js
```diff
@@ -6,7 +6,16 @@ describe('Evil promises should not be able to break invariants', function () {
...
     // Promise.resolve(evilPromise) is just the identity function.
-    var EvilPromise = function (executor) { Promise.call(this, executor); };
+    // (And in fact, most native Promise implementations use a private
+    // [[PromiseConstructor]] field in `Promise.resolve` which can't be
+    // easily patched in an ES5 engine, so instead of
+    // `Promise.resolve(evilPromise)` we'll use
+    // `new Promise(function(r){r(evilPromise);})` below.)
+    var EvilPromise = function (executor) {
+      var self = new Promise(executor);
+      Object.setPrototypeOf(self, EvilPromise.prototype);
```
ljharb: Please make sure this test bails out early if `Object.setPrototypeOf` is not shimmed in the current environment.

test/promise/simple.js
```diff
@@ -49,4 +49,48 @@ describe('Promise', function () {
       Promise.all.call(StealingPromiseConstructor, iterable);
     });
   });
+
+  specify('resolve with a thenable calls it once', function () {
```
ljharb: Both of these are an async test, so it needs the `done` argument - ie, `Promise.all([a, b]).then(done, done)` at the bottom of the tests.

ljharb:
Thanks for doing this! I think the "default species" stuff, and the "Array.from" stuff, might work better as separate PRs first, to keep this one smaller.

es6-shim.js
cscott: Yeah, my motivation for breaking it out is just (a) to mirror the spec more closely so you can read some of our implementations and match them step-by-step to the spec, and (b) to provide a central place to put special cases for IsConstructor (like arrow functions, or whatever) when we find a good reason to do so (for compatibility in some edge case).  But it doesn't seem worth while to start cluttering up IsConstructor yet until we have the use case for it.

ljharb: Agreed, this is strictly an improvement as-is

es6-shim.js
cscott: GetMethod doesn't complain if the property is missing, just if it's present but not callable.  As the comment inside that check indicates, the explicit check gives us better error messages in case the property is missing or null (in which case ES.GetMethod returns null or undefined).

It's not a correctness issue, since (as in the spec) the `_call` immediately following will throw the appropriate TypeError if `itFn` is null or undefined.  But you'd be looking at a pretty opaque TypeError in that case, just complaining that "undefined is not a function" or something like that.  This test lets us give a more friendly type error.

es6-shim.js
cscott: Done.

es6-shim.js
cscott: I expect this to be very hot code, since it will be executed every time we create a Map, Set, or Promise.  I've gotten complaints that our implementations were slow to construct in the past.  The `for...in` loop ought to be optimized better than `Object.keys(...).forEach(function() { .... })` (which requires the creation of an extra array and function environment).  We don't really need to worry about obscure bugs because the `slots` are always coming from our code, this parameter isn't exposed to users of the shim.

es6-shim.js
cscott: Updating the iterator helpers to match the spec requires changing Promise.race, Promise.all, and Array.from -- those are the three places where iterators are used.  It's very awkward to change one without changing the others.

test/promise/all.js
cscott: Already done, two lines below your note.

test/promise/evil-promises.js
cscott: Already done, two lines below your note.

test/promise/simple.js
cscott: No, mocha supports returning a Promise as the result of a test.  This is a much more convenient mechanism for async tests, and fixes bugs in your original implementation.

cscott:
Splitting the patch would require me to fixup the old implementation of Promises, just before I got rid of it.  I'd prefer not to waste the effort, assuming you can manage reviewing this as is.

es6-shim.js
ljharb: Sweet, I'm convinced

ljharb:
Hm, it seems like you could leave the `Promise` implementation and `ES.Construct` alone in the first PR, and get in the `addDefaultSpecies` and new `ES` methods separately? If you don't think it would add much then that's fine.

ljharb:
(also it looks like the tests are failing because we're not clobbering the native `Promise` when some of the new tests fail)

test/promise/simple.js
ljharb: That's super fancy, thanks for explaining

cscott:
Updated the patch to fix the wrapping-paren issue, and added a test on Promise.resolve() which should correct the test failures.

test/promise/evil-promises.js
```diff
@@ -20,7 +29,7 @@ describe('Evil promises should not be able to break invariants', function () {
     };
 
     var calledAlready = false;
-    Promise.resolve(evilPromise).then(function (value) {
+    new Promise(function (r) { r(evilPromise); }).then(function (value) {
```
ljharb: Hmm - `Promise.resolve` should have the same behavior as using `resolve` from `new Promise`, shouldn't it?

cscott: `Promise.resolve` has this sorta-sketchy "is the argument of the same class as `this`"  shortcut, which is what the `new Promise` version is explicitly avoiding.

ljharb:
Thanks for bearing with me - this looks awesome, much appreciated

PR #352, opened 2015-06-17
Update `Promise.reject` to match official ECMA-262 spec.

cscott:
There was a last-minute change to `Promise.reject` to make it better match `Promise.resolve`.  See:
http://www.ecma-international.org/ecma-262/6.0/#sec-promise.reject


ljharb:
I'll merge this now, but can you follow up with a test that covers this?

PR #359, opened 2015-08-19
Fix for IE11 check before `preserveToString`

AlastairTaft:
When in IE11 the 'preserve string' method crashes on `startsWith`. `startsWith` doesn't exist in IE11 so the `preserveToString` method doesn't work (there's nothing to preserve).

Added a check to prevent it running if it doesn't need to and it's going to cause an error.


ljharb:
Thanks for the report!

I'd rather fix this in the place `overrideNative` is being called on a non-present native. It seems like if [this](https://github.com/paulmillr/es6-shim/blob/master/es6-shim.js#L735-L739) checked that `startsWith` and `endsWith` both exist, before entering the block, it might solve it also?

That said, which version of IE 11 are you using? In my copy, this does not error out.

AlastairTaft:
In the `overrideNative` method you might be relying on the bit of code `defineProperty(object, property, replacement, true);` to set and make the `startsWith` property exist, but I'm not sure with this.

I think the `startsWithIsCompliant` is being set to `undefined` because it looks like both methods `startsWithRejectsRegex` and `startsWithHandlesInfinity` are probably returning `undefined` because they have checks like this at the start `return String.prototype.startsWith && ...`. So the setting of the value will be `startsWithRejectsRegex() && startsWithHandlesInfinity` will be `undefined && undefined`.

My IE version says
Version: 11.0.9600.17937 
UpdateVersion:11.0.22

ljharb:
I think you're totally correct that I'm incorrectly calling `overrideNative` in this case - I'm just surprised it hasn't failed before.

If you're open to editing and rebasing your PR so that it avoids https://github.com/paulmillr/es6-shim/blob/master/es6-shim.js#L735-L739 entirely when startsWith or endsWith aren't present, I'd be happy to merge that! As is, I'd prefer `overrideNative` to blow up when passed a nonexistent native, since it indicates that I did something wrong :-)

es6-shim.js
```diff
@@ -156,7 +156,9 @@
   var globals = getGlobal();
   var globalIsFinite = globals.isFinite;
   var hasStrictMode = (function () { return this === null; }.call(null));
-  var startsWithIsCompliant = startsWithRejectsRegex() && startsWithHandlesInfinity;
+  var startsWithExistAndIsNotCompliant = 
+    (typeof String.prototype.startsWith !== 'undefined') && 
+    (!startsWithRejectsRegex() || !startsWithHandlesInfinity());
```
ljharb: Please put all this on one line - line length limits are just as silly as arbitrary line breaks to accommodate them :-)

es6-shim.js
```diff
@@ -265,8 +267,12 @@
 
   var overrideNative = function overrideNative(object, property, replacement) {
     var original = object[property];
+    if (typeof original == 'undefined'){
+      throw "Native property does not exist";
```
ljharb: Please never ever throw anything but a `new Error` object - `TypeError` works here.

es6-shim.js
```diff
@@ -265,8 +267,12 @@
 
   var overrideNative = function overrideNative(object, property, replacement) {
     var original = object[property];
+    if (typeof original == 'undefined'){
```
ljharb: Please avoid `==`, `===` is needed here. Also, please put a space between `)` and `{`.

ljharb:
Awesome! If you can fix these comments, and then rebase it down to a single commit on top of the latest master, this will be good to go!

ljharb:
Please be sure to rebase this - merge commits make git logs sad

AlastairTaft:
My git skills are definitely lacking because I can't seem to squash these commits.

ljharb:
@AlastairTaft `git fetch ; git rebase -i origin/master` and edit the messages according to the instructions, and when you're done, `git push -f`

PR #366, opened 2015-10-31
Add tests for Number constructor that highlight deficiencies in certain environments.

Xotic750:
These were recently fixed in lodash trunk, _.toNumber.

Original PR https://github.com/lodash/lodash/pull/1577
Actual merge https://github.com/lodash/lodash/commit/37955345ef3d515e9c2799b82fad545a0c22efce

### DIFF
--- a/test/number.js
+++ b/test/number.js
@@ -330,7 +330,7 @@ describe('Number', function () {
   describe('strings in the constructor', function () {
     it('works on normal literals', function () {
       expect(Number('1')).to.equal(+'1');
-	  expect(Number('1.1')).to.equal(+'1.1');
+      expect(Number('1.1')).to.equal(+'1.1');
       expect(Number('0xA')).to.equal(0xA);
     });
   });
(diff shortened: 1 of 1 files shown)


Xotic750:
There are a couple of others that I will try to add.

test/number.js
```diff
@@ -387,5 +387,31 @@ describe('Number', function () {
       expect(Number('0o11')).to.equal(9);
       expect(Number({ toString: function () { return '0o12'; }, valueOf: function () { return '0o13'; } })).to.equal(11);
     });
+
+    it('should produce NaN', function () {
```
ljharb: these tests are failing but it's very hard to figure out which line is failing.

Perhaps `expect(String(Number(…))).to.equal('NaN')` would be better? I don't think we need to worry about testing that `Number()` returns a number :-)

Xotic750:
Changed those to String and 'NaN', hope that helps you a little more.

ljharb:
Thanks, this is highlighting a real failure. Can you please rebase these two commits down to one? I'll merge them manually, along with the fix.

Xotic750:
There are some other tests that highlight other issues but I guess we can take those separately?

ljharb:
I can leave this PR open, and you can keep adding commits to it, but if you rebase the existing diff in a single commit, I can cherry-pick it into master.

Xotic750:
I can do that when I have a little time (next week), I think some of this affects your https://github.com/ljharb/es-abstract and https://github.com/ljharb/es-to-primitive projects.

ljharb:
It definitely does - please open issues/PRs there as well when you can :-)

ljharb:
Turns out `to-primitive` doesn't have the bug, as it doesn't apply there. `es-abstract` is now fixed. Thanks!

Xotic750:
No problem, I think you still have the badHex issue though, some older environments, node v0.8 accepts +/-Hex value. But you may not be worried about such environments.

ljharb:
@Xotic750 I am indeed worried about those environments - more tests are welcome.

PR #368, opened 2015-11-02
Add tests to highlight Number constructor whitspace handling deficiencies

Xotic750:
```Number('  2  ') === 2```
So just like with trim, different environments handle internal trimming differently and inconsistently.


ljharb:
Related to #354

ljharb:
Thanks, fixed.

PR #369, opened 2015-11-04
Fix for bad hex strings when using `Number` constructor.

Xotic750:
Ref https://github.com/paulmillr/es6-shim/pull/366

Node v0.8, old Chrome/Chromium and others.

### DIFF
--- a/es6-shim.js
+++ b/es6-shim.js
@@ -719,6 +719,7 @@
   };
   var nonWS = ['\u0085', '\u200b', '\ufffe'].join('');
   var nonWSregex = new RegExp('[' + nonWS + ']', 'g');
+  var isBadHexRegex = /^[\-+]0x[0-9a-f]+$/i;
   var hasStringTrimBug = nonWS.trim().length !== nonWS.length;
   defineProperty(String.prototype, 'trim', trimShim, hasStringTrimBug);
 
@@ -1172,6 +1173,7 @@
       throw new TypeError('No default value');
     };
     var hasNonWS = nonWSregex.test.bind(nonWSregex);
+    var isBadHex = isBadHexRegex.test.bind(isBadHexRegex);
     var NumberShim = (function () {
       // this is wrapped in an IIFE because of IE 6-8's wacky scoping issues with named function expressions.
       var NumberShim = function Number(value) {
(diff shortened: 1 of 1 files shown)


ljharb:
Can you please provide tests that would fail without your change?

Also, can you link to spec text that says that `'-0x1'` should be `NaN`, when the literal `-0x1` parses fine?

Xotic750:
I already provided the tests in https://github.com/paulmillr/es6-shim/pull/366 and that was the only failing test remaining in your latest push.

ljharb:
You're right about the tests - https://travis-ci.org/paulmillr/es6-shim/jobs/89172994 passes but https://travis-ci.org/paulmillr/es6-shim/jobs/88743727 fails.

It's worth noting that this appears to make things consistent, but I do want to know what the spec requires.

Xotic750:
i would guess these
http://www.ecma-international.org/ecma-262/6.0/#sec-tonumber-applied-to-the-string-type
> StrNumericLiteral :::
    StrDecimalLiteral
    BinaryIntegerLiteral
    OctalIntegerLiteral
    HexIntegerLiteral

http://www.ecma-international.org/ecma-262/6.0/#sec-number-conversions

> HexIntegerLiteral :::
    0x HexDigit
   0X HexDigit
   HexIntegerLiteral HexDigit

So, current environments seem to interpret this as `'0x1'` is valid but `'-0x1'` is invalid.

es6-shim.js
```diff
@@ -719,6 +719,7 @@
   };
   var nonWS = ['\u0085', '\u200b', '\ufffe'].join('');
   var nonWSregex = new RegExp('[' + nonWS + ']', 'g');
+  var isBadHexRegex = new RegExp('^[-+]0x[0-9a-f]+$', 'i');
```
ljharb: This can just be a literal - `/^[\-+]0x[0-9a-f]+$/i`

Xotic750: I wasn't sure on the preference of style.

ljharb: In every language, using literals is preferred over more verbose forms when they're equivalent :-)

ljharb:
Please file a separate issue for `parseInt`

es6-shim.js
Xotic750: Understood, but I seem to remember some older environments had problems with certain literals, I think it was `XRegExp` that pointed these out. :)

PR #377, opened 2015-11-29
Not entry object and SameValueZero

Xotic750:
I was working on my ES3 compatibility shims (non-intrusive) and I wanted to use native Map and Set where available. Now, I know there have been issues in environments and so I thought I would use es6-shim as the tests seem to show that the issues have been resolved. To my surprise I found that some of my tests began to fail. I have added the tests to this PR so that you can see, either there is something wrong with my tests (but my compatibility shims pass all of these), or these are native bugs that have not yet been picked up, or there is something wrong with the es6-shims/fixes. I haven't investigated too far.

On node v0.10
```
[graham@tyr es6-shim]$ node
> new Map([1]);
ReferenceError: Map is not defined
    at repl:1:5
    at REPLServer.self.eval (repl.js:110:21)
    at repl.js:249:20
    at REPLServer.self.eval (repl.js:122:7)
    at Interface.<anonymous> (repl.js:239:12)
    at Interface.EventEmitter.emit (events.js:95:17)
    at Interface._onLine (readline.js:202:10)
    at Interface._line (readline.js:531:8)
    at Interface._ttyWrite (readline.js:760:14)
    at ReadStream.onkeypress (readline.js:99:10)
> require("./es6-shim.js");
> new Map("1");
{}
> new Map([1]);
{}
```
So the shim didn't throw a TypeError when I expected it to throw.

```
try {
    new Map([1]);
} catch (e) {
    console.log('threw', e);
}
try {
    new Map('1');
} catch (e) {
    console.log('threw', e);
}
```
http://jsfiddle.net/Xotic750/qoswzuye/

With Chromium 46.0.2490.80
```
threw TypeError: Iterator value 1 is not an entry object(…)
threw TypeError: Iterator value 1 is not an entry object(…)
```


test/collections.js
```diff
@@ -123,6 +123,11 @@ describe('Collections', function () {
       expect(map2).to.have.entries([['a', 'b'], ['c', 'd']]);
     });
 
+    it('non-entry objects should throw', function () {
+      expect(new Map([1, 2, 3])).to['throw'](TypeError);
```
ljharb: According to the [spec](http://tc39.github.io/ecma262/#sec-map-iterable), step 8, it simply calls `[0]` and `[1]` on the thing the iterable returns. In other words, this should be an iterable of three `undefined → undefined` entry pairs.

test/collections.js
```diff
@@ -123,6 +123,11 @@ describe('Collections', function () {
...
+      expect(new Map('123')).to['throw'](TypeError);
```
ljharb: This should be the same as `new Map(['1', undefined], ['2', undefined], ['3', undefined])`

ljharb:
I've commented on your first two tests, which I think aren't correct - I honestly have no idea what the rest of the tests are testing, nor what else specifically fails.

test/collections.js
ljharb: It looks like, for example, Safari's native implementation does throw here - I'll confirm what the correct behavior should be, and then either patch it or fix the shim.

ljharb: ok - so the specific step `If Type(nextItem) is not Object` is what gets triggered here, in both cases, because strings and numbers are primitives. However, `[{}, {}, {}]` would be valid, as would `[[], [], []]`. I'll add some tests, and then patch them.

ljharb:
Thanks as always for the report! I've fixed the "iterable" cases. However, it looks like this is multiple issues - would you mind filing a new one for the SameValueZero stuff?

Xotic750:
Yes, it was two issues. The SameValueZero one seems weird as you already have tests for it (which pass), so my guess is that it is something specific about the order of creation/replacement or something, again I haven't investigated. I will open another issue/PR.

PR #378, opened 2015-11-29
Map: Same value zero failing.

Xotic750:
Here  are the tests that show up the SameValueZero issue. All of the tests are required for it to happen, if you strip out what you may think are not necessary then you will get a pass, just like with your original tests. So, it is something to do with the particular order of happenings.

https://travis-ci.org/paulmillr/es6-shim/builds/93772657

My guess is that it is an JS engine bug as Node v0.10 passes, which is using the es6-shim shim. But it could be a bug in a patch.


ljharb:
This is definitely a weird v8 engine bug. Not all of the lines are required, but it seems to happen specifically when the Map's `size` is `4` before the `-0` is set.

ljharb:
Thanks for the report!

PR #379, opened 2015-12-03
Add missing checks to Promise.resolve and Promise.reject.

cscott:
These checks were overlooked when @@species was resolved from these methods.

Discovered by running the ES2015 test262 test suite against es6-shim.

### DIFF
--- a/es6-shim.js
+++ b/es6-shim.js
@@ -2316,6 +2316,9 @@
 
       reject: function reject(reason) {
         var C = this;
+        if (!ES.TypeIsObject(C)) {
+          throw new TypeError('Bad promise constructor');
+        }
         var capability = new PromiseCapability(C);
         var rejectFunc = capability.reject;
         rejectFunc(reason); // call with this===undefined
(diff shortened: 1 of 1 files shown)


ljharb:
Ah, thanks - could we add test coverage for these to this repo also?

ljharb:
These checks are already covered by `IsConstructor` inside `PromiseCapability` - I added https://github.com/paulmillr/es6-shim/commit/3d5bc93597ce89eb05630bcbe194cbf8e8ce3383 to be sure.

Can you share the test262 test conditions that cause a failure without this PR's changes?

cscott:
It's the `built-ins/Promise/resolve/context-non-object-with-promise.js` test case which is failing.  It passes `undefined` as the context object, but a valid `Promise` as the argument, so it never gets to the check in `IsConstructor`.  I added PR#381 to demonstrate how test262 test cases could be run on es6-shim, which should let you reproduce the failure.

ljharb:
The only difference I see in the test262 tests is that they're voiding the `constructor` property on the promise they pass in - however, looking at es6-shim's code path for resolve/reject, they both immediately do effectively `new PromiseCapability(this)`, which in its first line throws if `IsConstructor` fails for the argument, which it would for any non-function "this" value.

I definitely see the test failing in #381 but I'm mystified as to why.

ljharb:
aha, although `Promise.reject` already works fine, `Promise.resolve` has an `ES.IsPromise` check inside it.

ljharb:
@cscott Why does resolve have that check but not reject?

ljharb:
I'm pursuing that question with the committee members, but in the meantime, if you could freshly rebase this on top of latest master I'd love to merge it ASAP. Thanks!

cscott:
Rebased.

ljharb:
Thanks!

PR #381, opened 2015-12-12
Add test262 test cases.

cscott:
This currently only works on node, and only runs Promise-related test cases.

It could be generalized to run a larger subset of the ES2015 test cases, and to not fail horribly when run in a browser.


ljharb:
I'd love to add all of test262's relevant test cases - but I'm skeptical about "not [failing] horribly when run in a browser" :-)

cscott:
Well, since the browser doesn't have access to npm, you'd either have to (a) make the test262 package accessible somehow from within the browser, or (b) run some pre-processing step to turn all the test262 test cases into individual mocha tests inside the `test` directory.  Option (b) is probably not too hard, actually.

The other issue with running more of test262 is that they expect each test to have an isolated environment -- ie, you can stomp on a global, and then when you're running the next test things should go back to normal.  Mocha doesn't provide that by default.  I've deal with this so far by blacklisting the Promise-related tests which stomp on globals; it would be nice to have a better solution.

cscott:
I should have said that option (c) is just to do a test for `typeof process === 'undefined'` and if so, bail out of the test case.  That would just skip those tests when run in a browser, which would be the minimum thing necessary to "avoid failing horribly".

PR #383, opened 2015-12-12
Improve performance of Promises

cscott:
These patches improve the performance of Promises by roughly a factor of 2x, resulting in performance on the `doxbee sequential` benchmark equivalent to the `core-js` Promise implementation and within a factor of two of v8's native Promise implementation.  It also results in about 50% less memory usage on that benchmark than either `core-js` or v8 native Promises.

The improvements to TypeIsObject and IsCallable also improve the performance of other parts of the shims.

### DIFF
--- a/es6-shim.js
+++ b/es6-shim.js
@@ -2069,8 +2069,9 @@
       // global Promise below (in order to workaround bugs)
       // https://github.com/Raynos/observ-hash/issues/2#issuecomment-35857671
       var P = globals.Promise;
-      return P && P.resolve && function (task) {
-        return P.resolve().then(task);
+      var pr = P && P.resolve && P.resolve();
+      return pr && function (task) {
+        return pr.then(task);
       };
     };
     /*global process */
--- a/es6-shim.js
+++ b/es6-shim.js
@@ -286,10 +286,9 @@
     },
 
     TypeIsObject: function (x) {
-      /* jshint eqnull:true */
-      // this is expensive when it returns false; use this function
-      // when you expect it to return true in the common case.
-      return x != null && Object(x) === x;
+      return x !== void 0 && x !== null && x !== true && x !== false &&
+        typeof x !== 'string' && typeof x !== 'number' &&
+        typeof x !== 'symbol';
     },
 
     ToObject: function (o, optMessage) {
--- a/es6-shim.js
+++ b/es6-shim.js
@@ -2096,6 +2096,8 @@
     var PROMISE_FULFILL_OFFSET = 0;
     var PROMISE_REJECT_OFFSET = 1;
     var PROMISE_CAPABILITY_OFFSET = 2;
+    // This is used in an optimization for chaining promises via then.
+    var PROMISE_FAKE_CAPABILITY = {};
 
     var enqueuePromiseReactionJob = function (handler, capability, argument) {
       enqueue(function () {
@@ -2105,6 +2107,11 @@
 
     var promiseReactionJob = function (handler, promiseCapability, argument) {
       var handlerResult, f;
+      if (promiseCapability === PROMISE_FAKE_CAPABILITY) {
+        // Fast case, when we don't actually need to chain through to a
+        // (real) promiseCapability.
+        return handler(argument);
+      }
       try {
         handlerResult = handler(argument);
         f = promiseCapability.resolve;
@@ -2207,18 +2214,29 @@
       return { resolve: resolve, reject: reject };
     };
 
+    var optimizedThen = function (then, thenable, resolve, reject) {
+      // Optimization: since we discard the result, we can pass our
+      // own then implementation a special hint to let it know it
+      // doesn't have to create it.  (The PROMISE_FAKE_CAPABILITY
+      // object is local to this implementation and unforgeable outside.)
+      if (then === Promise$prototype$then) {
+        _call(then, thenable, resolve, reject, PROMISE_FAKE_CAPABILITY);
+      } else {
+        _call(then, thenable, resolve, reject);
+      }
+    };
     var promiseResolveThenableJob = function (promise, thenable, then) {
       var resolvingFunctions = createResolvingFunctions(promise);
       var resolve = resolvingFunctions.resolve;
       var reject = resolvingFunctions.reject;
       try {
-        _call(then, thenable, resolve, reject);
+        optimizedThen(then, thenable, resolve, reject);
       } catch (e) {
         reject(e);
       }
     };
 
-    var Promise$prototype;
+    var Promise$prototype, Promise$prototype$then;
     var Promise = (function () {
       var PromiseShim = function Promise(resolver) {
         if (!(this instanceof PromiseShim)) {
@@ -2294,7 +2312,7 @@
           index, values, resultCapability, remaining
         );
         remaining.count += 1;
-        nextPromise.then(resolveElement, resultCapability.reject);
+        optimizedThen(nextPromise.then, nextPromise, resolveElement, resultCapability.reject);
         index += 1;
       }
       if ((--remaining.count) === 0) {
@@ -2323,7 +2341,7 @@
           throw e;
         }
         nextPromise = C.resolve(nextValue);
-        nextPromise.then(resultCapability.resolve, resultCapability.reject);
+        optimizedThen(nextPromise.then, nextPromise, resultCapability.resolve, resultCapability.reject);
       }
       return resultCapability.promise;
     };
@@ -2418,7 +2436,13 @@
         var promise = this;
         if (!ES.IsPromise(promise)) { throw new TypeError('not a promise'); }
         var C = ES.SpeciesConstructor(promise, Promise);
-        var resultCapability = new PromiseCapability(C);
+        var resultCapability;
+        var returnValueIsIgnored = (arguments[2] === PROMISE_FAKE_CAPABILITY);
+        if (returnValueIsIgnored && C === Promise) {
+          resultCapability = PROMISE_FAKE_CAPABILITY;
+        } else {
+          resultCapability = new PromiseCapability(C);
+        }
         // PerformPromiseThen(promise, onFulfilled, onRejected, resultCapability)
         // Note that we've split the 'reaction' object into its two
         // components, "capabilities" and "handler"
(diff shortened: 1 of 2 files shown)


es6-shim.js
```diff
@@ -291,6 +291,13 @@
       return x != null && Object(x) === x;
     },
 
+    TypeIsNotObject: function (val) {
+      // This is faster when TypeIsObject may commonly return false.
+      return val === void 0 || val === null || val === true || val === false ||
+        typeof val === 'string' || typeof val === 'number' ||
+        typeof val === 'symbol'; // jshint ignore:line
```
ljharb: `symbol` is a valid typeof value, so rather than the jshint comment, let's modify the project config to not complain about that.

es6-shim.js
```diff
@@ -291,6 +291,13 @@
       return x != null && Object(x) === x;
     },
 
+    TypeIsNotObject: function (val) {
+      // This is faster when TypeIsObject may commonly return false.
+      return val === void 0 || val === null || val === true || val === false ||
```
ljharb: i think `return val == null || (typeof val === 'object' || typeof val === 'function')` is the simplest here?

es6-shim.js
```diff
@@ -291,6 +291,13 @@
       return x != null && Object(x) === x;
     },
 
+    TypeIsNotObject: function (val) {
```
ljharb: Let's name this `IsPrimitive`, since that will correspond to an actual Abstract Operation

es6-shim.js
```diff
@@ -2009,6 +2016,9 @@
         capability.resolve = resolve;
         capability.reject = reject;
       };
+      // Initialize fields to inform optimizers about the object shape.
```
ljharb: shouldn't they be initialized to a noop function instead then, so that the type of the value doesn't change?

es6-shim.js
```diff
@@ -2060,59 +2071,99 @@
...
+        _promise.rejectReactions0 = void 0;
+        _promise.reactionCapability0 = void 0;
+        if (length > 1) {
+          for (var i = 1, idx = 0; i < length; i++, idx += 3) {
+            enqueuePromiseReactionJob(
+              _promise[idx + PROMISE_FULFILL_OFFSET],
+              _promise[idx + PROMISE_CAPABILITY_OFFSET],
+              value
+            );
+            promise[idx + PROMISE_FULFILL_OFFSET] = void 0;
```
ljharb: similarly, if these will be numbers, should they be initialized to number values?

es6-shim.js
```diff
@@ -2147,18 +2198,29 @@
       return { resolve: resolve, reject: reject };
     };
 
+    var optimizedThen = function (then, thenable, resolve, reject) {
+      // Optimization: since we discard the result, we can pass our
+      // own then implementation a special hint to let it know it
+      // doesn't have to create it.  (The PROMISE_FAKE_CAPABILITY
+      // object is local to this implementation and unforgable outside.)
```
ljharb: s/unforgable/unforgeable

es6-shim.js
```diff
@@ -2227,7 +2296,8 @@
           index, values, resultCapability, remaining
         );
         remaining.count += 1;
-        nextPromise.then(resolveElement, resultCapability.reject);
+        optimizedThen(nextPromise.then, nextPromise,
+                      resolveElement, resultCapability.reject);
```
ljharb: let's keep this on one line - arbitrary line length limits hurt readability imo

es6-shim.js
```diff
@@ -2045,14 +2055,15 @@
...
+      var pr = P && P.resolve && P.resolve();
+      return pr && function (task) {
+        return pr.then(task);
       };
     };
     /*global process */
     /* jscs:disable disallowMultiLineTernary */
     var enqueue = ES.IsCallable(globals.setImmediate) ?
-      globals.setImmediate.bind(globals) :
+      globals.setImmediate :
```
ljharb: good catch, i guess this bind was just never necessary?

ljharb:
Thanks, this looks great! I have a number of minor comments but after those and a fresh rebase, this should be good to go.

es6-shim.js
ljharb: i think `return val == null || (typeof val !== 'object' && typeof val !== 'function')` is the simplest here?

es6-shim.js
cscott: That doesn't seem to be a problem when I look at the generated optimized assembly.

And it's a little nontrivial, since the spec explicitly says the type should be undefined, that is, it allows you to call `resolver(undefined, undefined)` before actually passing the real resolver functions and it should be a no-op.  So the type check in line 2006/2013 above would have to be something like:
`if ((capability.resolve !== void 0 && capability.resolve !== NOOP_FUNC) || ...`
and the extra complexity isn't worth it -- again, since looking at the optimized assembly seems to indicate it doesn't matter.

es6-shim.js
cscott: Yup, as far as I can tell.  It was just "extra safety" -- but that extra safety really slows things down.

es6-shim.js
cscott: They are not numbers, and this is after they are being used in any case, just to free them. In the optimizer it's *runtime* values which are important, and no one will touch these fields after they are set to undefined, so the optimizer will never take the bailout to recompile them with new types.

es6-shim.js
cscott: http://www.ecma-international.org/ecma-262/6.0/#sec-ecmascript-data-types-and-values actually only defines the `Type` operation and then the spec says things like `if Type(x) is Object, then..`.  There is no mention of `IsPrimitive` in the spec text.

On the other hand, the `TypeIsNotObject` form actually seems to be faster even when the argument is an Object.  So perhaps I should just combine these two cases and use the `TypeIsNotObject` implementation always.  I'd originally preferred the `TypeIsObject` version because it corresponded more closely to the spec text, but that correspondence seems to be gone in the latest spec version.

es6-shim.js
cscott: done.

es6-shim.js
cscott: done

es6-shim.js
cscott: done

es6-shim.js
cscott: `val == null` is actually slower than an explicit test against `null` and `undefined`, when you look at v8's optimized assembly output.  The form shown here was pretty extensively benchmarked.  It might look a little weird, but it's pretty straightforward to read (a plus) and happens to be the fastest option.

cscott:
Ok, rebased with fixes from your comments applied.

es6-shim.js
ljharb: k, if it doesn't make a difference, that's just fine then

es6-shim.js
ljharb: sounds good

ljharb:
@cscott crap, sorry, I cherry picked two of your commits and pushed them up not realizing you'd already done the rebase. Would you mind one more rebase? :-) if not i'll merge manually

es6-shim.js
ljharb: What about `return val === void 0 || val === null || (typeof val !== 'object' && typeof val !== 'function')`? How much faster are we talking about?

cscott:
rebased, no problem.

es6-shim.js
ljharb: my main concern is that new primitive typeof values (which there will likely be many) will require a whitelist approach to be continually updated, whereas a blacklist approach will work essentially forever.

cscott: I think the key to the optimization is that `undefined`, `null`, `true`, `false` etc are all actually constant pointers which can be compared directly without dereferencing.  Similarly `number` is a tagged value (a tagged NaN) so it can also be a direct comparison.   That makes those tests extremely fast if they return true.  We only have to actually dereference for `string` and `symbol` (and I might be wrong about those, likely there are some optimizations for string) -- which is equivalent to the two dereferences needed for `object` and `function` in your case, except you've lost the fast cases for booleans and numbers.

Of course, this is only if the argument is completely unknown.  If the function is inlined with a known type this compiles out to a constant in any case.

(There might be a special weirdness about `typeof val === 'function'` -- when looking at generated code, v8 doesn't seem to do the same type-tracking of function types that it does for object types.  I don't know if that actually comes into play in this case or not.)

cscott: > my main concern is that new primitive typeof values (which there will likely be many) will require a whitelist approach to be continually updated, whereas a blacklist approach will work essentially forever.

Actually the opposite -- see the last case in http://www.ecma-international.org/ecma-262/6.0/#sec-typeof-operator-runtime-semantics-evaluation .  New exotic objects will not require this explicit list to be redefined.

Of course, if a new *primitive* type is added (like `Symbol`), all bets are off.  But not much we can do about that, other than update es6-shim when that day comes.

ljharb: SIMD is a ton of new primitive types, and Value Types will be a bunch of new ones as well. That's my concern - I'd rather not have to update es6-shim to make these new additions work.

cscott: It's a spec-correctness issue, see the link above.

ljharb: right - but this table *will, guaranteed* be extended in the future to add more primitive types. SIMD is stage 3, soon to be stage 4, and defines like 20 of them. As-is, that means `es6-shim` will have to add special cases for all SIMD values, and any intervening versions that contain this change will be broken in future implementations that expose the new primitives.

cscott: From reading
http://comments.gmane.org/gmane.comp.lang.javascript.ecmascript4.general/34503
and http://dhtmlkitchen.com/exotic-objects/ it seems that most exotic
(host) objects actually return "object" or "function".  I couldn't find an
example that returned something else but should still be considered an
object.

Perhaps the easiest thing to do is merge my PR and then do a follow-up
patch which replaces TypeIsObject with your preferred version?  I still
maintain it's not actually correct according to the spec, but I also agree
that adding a huge list of SIMD cases here would be very unpleasant.

ljharb: Sure, let's do that.

ljharb: Added the refactor here: 8dc13d74a5a2e360b21b76ce6c1bad13e6e76e19

The only "slowdown" should be in the case of when it *is* an object, in which case, it's two extra string comparisons.

cscott: Factoring out the `typeof x` might actually lead to a slowdown.  `typeof x == <string constant>` is optimized as a complete expression, breaking it up might defeat the optimization.  I'll check that.  (There are no actual string comparisons performed after the optimization.)

cscott: (Incidentally, http://mrale.ph/irhydra/2/ is a relatively-friendly tool that lets you see the compiled/optimized assembly.)

PR #388, opened 2015-12-16
Re-optimize `ES.TypeIsObject`.

cscott:
Caching the typeof value in a local variable defeats the `typeof x === <constant string>` peephole optimization, and slows down `es6-shim` on the doxbee benchmark (on my machine, 20000 iterations) from 5870ms (with this patch) to 6062ms (without it).

### DIFF
--- a/es6-shim.js
+++ b/es6-shim.js
@@ -285,15 +285,20 @@
       return x;
     },
 
+    // This might miss the "(non-standard exotic and does not implement
+    // [[Call]])" case from
+    // http://www.ecma-international.org/ecma-262/6.0/#sec-typeof-operator-runtime-semantics-evaluation
+    // but we can't find any evidence these objects exist in practice.
+    // If we find some in the future, you could test `Object(x) === x`,
+    // which is reliable according to
+    // http://www.ecma-international.org/ecma-262/6.0/#sec-toobject
+    // but is not well optimized by runtimes and creates an object
+    // whenever it returns false, and thus is very slow.
     TypeIsObject: function (x) {
       if (x === void 0 || x === null || x === true || x === false) {
         return false;
       }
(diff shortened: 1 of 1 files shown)


ljharb:
lulz @ js engine optimization

es6-shim.js
```diff
@@ -289,11 +289,16 @@
...
+      // This might miss the "(non-standard exotic and does not implement
+      // [[Call]])" case from
+      // http://www.ecma-international.org/ecma-262/6.0/#sec-typeof-operator-runtime-semantics-evaluation
+      // but we can't find any evidence these objects exist in practice.
+      // If we find some in the future, you could test `Object(x) === x`,
+      // which is reliable according to
+      // http://www.ecma-international.org/ecma-262/6.0/#sec-toobject
+      // but is not well optimized by runtimes and creates an object
+      // whenever it returns false, and thus is very slow.
+      return typeof x === 'function' || typeof x === 'object';
```
ljharb: fwiw i think we could do `|| Object(x) === x` here and only new primitives would hit the slow path, and only until we'd attempted to account for them - that might be better than leaving just the comment?

cscott: That's an interesting proposal.  You'd have to add back the string/number/symbol checks in that case, though, to ensure that only "exotics" hit the `|| Object(x)` case.  Not sure the extra complexity is worth it, given that I can't actually identify any valid "exotics" in any implementation that don't return 'object' or 'function'.  Presumably if there really is some instance in the wild someone will file a bug complaining that es6-shim throws a `TypeError` in some case when it shouldn't, and we can tweak the check then.

ljharb: ah yes, i missed that you removed those.

That's a fair point, so I think this is OK

PR #389, opened 2015-12-16
Further general optimizations.

cscott:
Be careful about how `arguments` is accessed, and ensure that the `ES.*` helpers are treated as constant functions which can be inlined.


es6-shim.js
```diff
@@ -267,7 +267,9 @@
 
   var $String = String;
 
-  var ES = {
+  // Assign these functions to a prototype to give v8 a hint that it
```
ljharb: how does this impact other engines? v8-specific optimizations have hurt other engines' performance in the past, with bluebird for example.

es6-shim.js
```diff
@@ -1096,7 +1098,9 @@
 
   var ArrayPrototypeShims = {
     copyWithin: function copyWithin(target, start) {
-      var end = arguments[2]; // copyWithin.length must be 2
+      // copyWithin.length must be 2, so we can't add `end` to the arguments
```
ljharb: this comment isn't necessary, we should definitely be consistently checking `arguments.length` before accessing an index - which we already do in many places.

es6-shim.js
```diff
@@ -2013,7 +2017,7 @@
     // some environments don't have setTimeout - no way to shim here.
     if (typeof setTimeout !== 'function' && typeof setTimeout !== 'object') { return; }
 
-    ES.IsPromise = function (promise) {
+    var IsPromise = function (promise) {
```
ljharb: why wouldn't we want this on the `ES` object?

es6-shim.js
```diff
@@ -2439,10 +2443,10 @@
 
       then: function then(onFulfilled, onRejected) {
         var promise = this;
-        if (!ES.IsPromise(promise)) { throw new TypeError('not a promise'); }
+        if (!IsPromise(promise)) { throw new TypeError('not a promise'); }
         var C = ES.SpeciesConstructor(promise, Promise);
         var resultCapability;
-        var returnValueIsIgnored = (arguments[2] === PROMISE_FAKE_CAPABILITY);
+        var returnValueIsIgnored = (arguments.length > 2 && arguments[2] === PROMISE_FAKE_CAPABILITY);
```
ljharb: might as well remove these unnecessary parens also

es6-shim.js
cscott: I can't imagine it would hurt performance, since it's just a throwaway assignment.  I'll try to double-check, but I don't know of any equivalent low-level IR tools on the firefox side.  (jsperf can be highly misleading.)

es6-shim.js
cscott: Well, I was just repeating and expanding the comment which was there originally, which was trying to explain why this couldn't be simply declared as `copyWithin(target, start, end)`.  But in theory our test cases should actually be verifying the length property of the function, so the comment should be unnecessary...

es6-shim.js
cscott: Because mutating an object after you've defined it ensures that `IsPromise` won't be optimized.  The number of slots is fixed at definition time.  But even if you add `IsPromise: null` to the definition of `ES` to hold the space, this ends up being initialized as a property reference stored in the ES object, not a "constant function reference" stored in the object's map (which can then be inlined after a hoisted type check).

This is described a bit by http://mrale.ph/blog/2012/09/23/grokking-v8-closures-for-fun.html

es6-shim.js
cscott: Hm, I guess---although I find having two different equals signs on a line without explicit grouping impairs readability.

ljharb:
I've cherry-picked in one of these commits - a rebase should result in only the one commit left on top

Issue #387, opened 2015-12-16
Map/Set iterator


Xotic750:
I am detecting `MapIterator` and `SetIterator` for my [inpect-x](https://github.com/Xotic750/inspect-x) module. On native ES6 `Map` and `Set`, the following code will alert `true` and `TypeError: Method Map Iterator.prototype.next called on incompatible receiver`.

```
var m = new Map();
var im = m.values();
var nm = im.next;
try {
  alert(nm.call(new Map([
    [1, true]
  ]).values()).value);
} catch (e) {
  alert(e);
}
try {
  alert(nm.call(new Set([true]).values()).value);
} catch (e) {
  alert(e);
}
```

http://jsfiddle.net/Xotic750/xw2Lpbwz/

But on `es6-shim` iterators it alerts `true` and `true`.

(I think it is the shim and not the native versions, but I have not tested thoroughly yet)

Safari8, FF26, IE10, node0.10 and 0.8 are examples.

I just did a `toString` of the `next` function and it is the shim version.

ljharb:
You'd need to do `Function.toString.call` to get the actual source of the function, just in case - the `es6-shim` definitely modifies the `toString` of many of its shims.

Xotic750:
Ok, but I didn't think that node 0.10 and 0.8 had native Map/Set.

ljharb:
They don't :-) that comes in 0.12.

Xotic750:
https://saucelabs.com/beta/tests/98e7c0128021479795a1c6cf0de06b46/watch

cscott:
The code in question seems a very strange way to detect the presence of `MapIterator` and `SetIterator`.  But regardless, #391 contains a fix.

Xotic750:
Thanks, I haven't tested it yet. The code I posted was just to demonstrate what I was seeing. The actual code is more like
```
var SET = typeof Set === 'function' && isSet(new Set()) && Set;
var testSet = SET && new SET(['SetSentinel']);
var sValues = SET && SET.prototype.values;

function isSetIterator(value) {
  if (!SET || !isObjectLike(value) || !ES.IsCallable(value.next)) {
    return false;
  }
  try {
    return ES.Call(
      value.next,
      ES.Call(sValues, testSet)
    ).value === 'SetSentinel';
  } catch (ignore) {}
  return false;
}
```
It's fairly expensive, I know, but accuracy is more important than performance for me and I couldn't figure out a better cross environment/realm way of doing it?

[...]

PR #391, opened 2015-12-16
Distinguish Set and Map iterators.

cscott:
Closes: #387

### DIFF
--- a/es6-shim.js
+++ b/es6-shim.js
@@ -2731,7 +2731,11 @@
         };
 
         MapIterator.prototype = {
+          isMapIterator: true,
           next: function next() {
+            if (!this.isMapIterator) {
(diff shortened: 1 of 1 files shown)


ljharb:
This seems fine, but we should probably have a test to prevent a regression

Xotic750:
What kind of tests would be needed for this?

ljharb:
@Xotic750 i'd assume something from your report in #387.

Although, I'd expect `Object.getPrototypeOf(new Map().entries()) !== Object.getPrototypeOf(new Set().entries())` to be `true`, and I'm not sure that's the case with the shims.

Xotic750:
https://github.com/paulmillr/es6-shim/pull/437

PR #397, opened 2016-01-10
Optimize Map/Set fast key path.

cscott:
The expression `typeof x === 'some string'` is treated as a peephole
optimization by JavaScript runtimes and optimized as a simple type test
without an explicit string comparison.  Refactoring the `typeof x` part
into a variable cause this to be deoptimized: the type information
is lost and we end up doing actual string comparison operations at
runtime.


ljharb:
Thanks!

PR #401, opened 2016-01-22
Remove special character and replace with regular apostrophe

cr-nz:
We have a build script which we are using but it doesn't accept special characters. I noticed it breaking on your project. The character was a `’` apostrophe and it's just in a comment. I have replaced that with a regular one.

### DIFF
--- a/es6-shim.js
+++ b/es6-shim.js
@@ -79,7 +79,7 @@
   };
 
   // Define configurable, writable and non-enumerable props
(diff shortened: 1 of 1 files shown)


ljharb:
This sounds like a flaw in your build script. Typographically correct apostrophes shouldn't cause problems, as the encoding of JS files should be assumed to be UTF-8 or UTF-16.

Issue #407, opened 2016-02-09
Uncaught (in promise) TypeError: object is not a function


akim-mcmath:
I'm getting this unhelpful error message in Chrome after upgrading to es6-shim@0.34.3.

```
Uncaught (in promise) TypeError: object is not a function            index.html:1
    at Object.defer (native)
    at chain (native)
    at then (native)
```

I'm using this as part of an Angular 2 application, bundled with Webpack. When I downgrade to es6-shim@0.34.2 the error goes away. And I'm not getting any errors in Firefox.

ljharb:
@akim-mcmath I assume you're using Chrome Canary?

This is a trivial console warning that you can absolutely ignore and won't affect your program in any way. That said, it should have been resolved by v0.34.3 (and was present in v0.34.2). If you could somehow narrow it down to a line in `es6-shim` that would help a lot.

fwiw, I can't reproduce this in Chrome or Chrome Canary with v0.34.3.

akim-mcmath:
@ljharb Thanks for your reply. I'm using Chrome 48.0.2564 actually. I also neglected to mention that the program seemed to be working fine otherwise. If it's trivial, then you don't need to worry about it for my sake. However, I seem to have found the source of the warning on [line 2562](https://github.com/paulmillr/es6-shim/blob/5f641326fe41fa85ccadb373a7e8c56c0acd0e2f/es6-shim.js#L2562) of v0.34.3. I don't know why it's not reproducible, but I believe I've found a fix, which I can submit if you would like.

ljharb:
Thanks, merged!

akim-mcmath:
You're welcome. Thank you for your work on this project and your quick resolution of this issue.

ljharb:
np! released as v0.34.4.

PR #408, opened 2016-02-09
Fix 'Uncaught (in promise) TypeError' in Chrome 48

akim-mcmath:
Fix for Issue #407.

On Chrome 48.0 in OSX, the following console warning started appearing with the release of es6-shim v0.34.3:

```
Uncaught (in promise) TypeError: object is not a function            index.html:1
    at Object.defer (native)
    at chain (native)
    at then (native)
```

The one-line change here fixes the warning. It seems that the warning was originating from the 'inner' promise `p` (the promise resolved by the 'outer' promise `p2`) on [line 2562](https://github.com/paulmillr/es6-shim/blob/5f641326fe41fa85ccadb373a7e8c56c0acd0e2f/es6-shim.js#L2562). The fix was to catch the rejection by promise `p` with a noop.

The warning is gone and all tests are passing.

### DIFF
--- a/es6-shim.js
+++ b/es6-shim.js
@@ -2561,7 +2561,7 @@
       p.constructor = {};
       var p2 = Promise.resolve(p);
       try {
-        p2.then(null, function () {}); // avoid "uncaught rejection" warnings in console
+        p2.then(null, noop).then(null, noop); // avoid "uncaught rejection" warnings in console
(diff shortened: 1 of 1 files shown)


ljharb:
This is super bonkers but I can confirm it fixes it. Thanks!

PR #412, opened 2016-03-01
Simplify logic for Math.tanh

benjoffe:
Previous PR for tanh left some code that was no longer needed (the checks against Infinity are redundant with the early exit).

### DIFF
--- a/es6-shim.js
+++ b/es6-shim.js
@@ -1960,11 +1960,8 @@
       // can exit early at +-20 as JS loses precision for true value at this integer
       if (x >= 20) { return 1; }
       if (x <= -20) { return -1; }
-      var a = Math.expm1(x);
-      var b = Math.expm1(-x);
-      if (a === Infinity) { return 1; }
-      if (b === Infinity) { return -1; }
-      return (a - b) / (Math.exp(x) + Math.exp(-x));
+
+      return (Math.expm1(x) - Math.expm1(-x)) / (Math.exp(x) + Math.exp(-x));
     },
 
     trunc: function trunc(value) {


es6-shim.js
```diff
@@ -1957,14 +1955,12 @@
...
       // can exit early at +-20 as JS loses precision for true value at this integer
       if (x >= 20) { return 1; }
       if (x <= -20) { return -1; }
-      var a = Math.expm1(x);
-      var b = Math.expm1(-x);
-      if (a === Infinity) { return 1; }
-      if (b === Infinity) { return -1; }
-      return (a - b) / (Math.exp(x) + Math.exp(-x));
+
+      return (Math.expm1(x) - Math.expm1(-x)) / (Math.exp(x) + Math.exp(-x));
```
benjoffe: Note: I not sure 4 math function calls really should be needed here, but I'll trust it has some accuracy benefit or something so not changing.

es6-shim.js
```diff
@@ -1939,9 +1939,7 @@
 
     sign: function sign(value) {
       var number = Number(value);
-      if (number === 0) { return number; }
-      if (Number.isNaN(number)) { return number; }
-      return number < 0 ? -1 : 1;
+      return number < 0 ? -1 : number > 0 ? 1 : number;
```
ljharb: here i kind of prefer the explicit steps - it makes it very clear what the edge cases are for `0` and `NaN`.

benjoffe: At the expense of performance? Would you be happy with a comment?

ljharb: What makes you think this has a meaningful performance impact? The only part that's not trivially optimizable is `Number.isNaN`, and that will be inlined by most every JIT.

benjoffe:
You're correct about neglibile performance difference for Math.sign changes, reverted.

ljharb:
Thanks, this looks great - if it's not too much to ask, could you rebase this on latest master?

Yaffle:
Another pull request contains a change for `Math.tanh`:
https://github.com/paulmillr/es6-shim/pull/338/files#diff-fbdfeb59ea7599799df83561186da7dfR1456

Yaffle:
```
Math.tanh = function (x) {
  return Math.sign(x) / (1 + 2 / Math.expm1(2 * Math.abs(x)));
};
```
What about this one?

ljharb:
@Yaffle i'd be happy to review a new PR that changed the implementation, with assorted tests (altho i'd prefer the edge cases be explicit)

PR #413, opened 2016-03-08
Math.atanh

Yaffle:


es6-shim.js
```diff
@@ -1844,13 +1844,7 @@
...
     atanh: function atanh(value) {
       var x = Number(value);
-      if (Number.isNaN(x) || x < -1 || x > 1) {
-        return NaN;
-      }
-      if (x === -1) { return -Infinity; }
-      if (x === 1) { return Infinity; }
-      if (x === 0) { return x; }
-      return 0.5 * _log((1 + x) / (1 - x));
+      return Math.sign(x) * Math.log1p(2 * Math.abs(x) / (1 - Math.abs(x))) / 2;
```
ljharb: please preserve the explicit edge case checks.

Yaffle: What is an edge case?
Why preserve?

ljharb: NaNs, zeroes, and infinities are edge cases. Preserving them makes it clear to future refactorers of these methods that they don't need to rely on things like `Math.sign` etc to handle this for them.

Also, we'll want to use cached versions of `Math.sign`, `Math.log1p`, and `Math.abs`, to avoid breaking if someone deletes them off of the `Math` object.

es6-shim.js
```diff
@@ -1850,7 +1850,7 @@
       if (x === -1) { return -Infinity; }
       if (x === 1) { return Infinity; }
       if (x === 0) { return x; }
-      return 0.5 * _log((1 + x) / (1 - x));
+      return Math.sign(x) * Math.log1p(2 * Math.abs(x) / (1 - Math.abs(x))) / 2;
```
ljharb: `_abs` is already available as a cached `Math.abs` - we should cache `sign` and `log1p` as well. (i'll make a cached `Math.sign` available on master shortly, so you can just rebase on top of it and use `_sign`)

es6-shim.js
Yaffle: @ljharb ,

```
var Math = {sign: OriginalMath.sign, log1p: OriginalMath.log1p, abs: OriginalMath.abs}; // "Cached" Math
```

ljharb: sure, that works too - it's just slower to have to do the internal object lookup, and shadowing the global makes it harder to shim it.

Yaffle: But there are some "pros" too, isn't it?

ljharb: only readability, but that's subjective, and i'd argue that using `Math.sign` and having it point not at the global `Math.sign` would make the code harder to understand - this way, grepping for `_sign` is pretty straightforward.

es6-shim.js
Yaffle: So `_log1p` and `_expm1` (to implement other things) are needed too

ljharb: yes, true - I'm ok with you punting on that for now tho, and i can take care of it all at once in a followup

Yaffle:
SHIT!

PR #425, opened 2016-09-28
Drop node support for 0.4 to 0.9

paladox:
Reason since most users are on node 0.10 or above and even then usage of 0.10 is going down.


ljharb:
Absolutely not - the whole point of a whim is to work, forever, on older engines.

Issue #422, opened 2016-08-09
set/get ngFor performance on IE11


aronallen:
Hi,

We are having this issue in Angular 2 land.
Any chance shim performance can be improved for IE11.

https://github.com/angular/angular/issues/8112

> The code is constantly calling isNan function in es6-shim.js.
> The problem seems to be due to the reason that set and get functions of the polyfill are extremely slow (compared to their native counterparts) and take most of the time.

ljharb:
I'd be happy to improve performance wherever it can be done without sacrificing correctness.

I'm not sure what the best way to do that is, but am open to PRs.

jdalton:
I'll give a PR a try :grinning:

ljharb:
@aronallen #429 is released with [v0.35.2](https://github.com/paulmillr/es6-shim/releases/tag/v0.35.2) - please let me know if that does not resolve your performance issue.

PR #429, opened 2016-11-21
Use original map for data backing when possible.

jdalton:
This PR attempts to address #422 by using the original map for data backing when possible to avoid linear lookups.

### DIFF
--- a/es6-shim.js
+++ b/es6-shim.js
@@ -205,6 +205,12 @@
   var ArrayIterator; // make our implementation private
   var noop = function () {};
 
+  var OrigMap = globals.Map;
+  var origMapDelete = OrigMap && OrigMap.prototype['delete'];
+  var origMapGet = OrigMap && OrigMap.prototype.get;
+  var origMapHas = OrigMap && OrigMap.prototype.has;
+  var origMapSet = OrigMap && OrigMap.prototype.set;
+
   var Symbol = globals.Symbol || {};
   var symbolSpecies = Symbol.species || '@@species';
 
@@ -2695,8 +2701,8 @@
 
   if (supportsDescriptors) {
 
-    var fastkey = function fastkey(key) {
-      if (!preservesInsertionOrder) {
+    var fastkey = function fastkey(key, skipInsertionOrderCheck) {
+      if (!skipInsertionOrderCheck && !preservesInsertionOrder) {
         return null;
       }
       if (isNullOrUndefined(key)) {
@@ -2868,8 +2874,9 @@
           var map = emulateES6construct(this, Map, Map$prototype, {
             _es6map: true,
             _head: null,
-            _storage: emptyObject(),
-            _size: 0
+            _map: OrigMap ? new OrigMap() : null,
+            _size: 0,
+            _storage: emptyObject()
           });
 
           var head = new MapEntry(null, null);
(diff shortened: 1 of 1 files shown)


ljharb:
Thanks!

What about browsers where `-0` and `0` are not properly treated? There's also [insertion order](https://github.com/paulmillr/es6-shim/blob/master/es6-shim.js#L2681-L2693), browsers where iteration completion [throws a StopIterator exception](https://github.com/paulmillr/es6-shim/blob/master/es6-shim.js#L3341-L3343), browsers where "size" [is a function](https://github.com/paulmillr/es6-shim/blob/master/es6-shim.js#L3345) instead of a getter, browsers where Map and Set [aren't iterable](https://github.com/paulmillr/es6-shim/blob/master/es6-shim.js#L3348) except with an eval'd `for-of`, browsers that lack the facility to [clear and create iterators for keys/values/entries](https://github.com/paulmillr/es6-shim/blob/master/es6-shim.js#L3346) and then combined with the fact that the shim itself mutates `Map.prototype` when it already exists - so you'd need to cache all of the existing prototype methods (like `get`, `has`, `set`, etc), or else you'd risk an infinite loop.

In other words, using the original Map is a great approach, but there are a fractal of issues caused by the very necessity of the shim in the first place :-/

jdalton:
> What about browsers where -0 and 0 are not properly treated?

I was hoping that was handled by the fast path. If not I'll rework to address it in the fast key path.

> There's also insertion order, browsers where iteration completion throws a StopIterator exception, 

Do you have a unit test to cover this? This is internal storage where the fast path has no concept of order since it's an empty object.

>  browsers where "size" is a function instead of a getter, browsers where Map and Set aren't iterable except with an eval'd for-of, 

I'm not using `size` of the internal map.

> browsers that lack the facility to clear and create iterators for keys/values/entries

I'm not sure that affects this internal use case.

> the shim itself mutates Map.prototype when it already exists - so you'd need to cache all of the existing prototype methods (like get, has, set, etc), or else you'd risk an infinite loop.

Ya, I'll address that.

ljharb:
I have a number of test cases that should cover it - you'd need to run the test HTML in a very large number of browsers (which sadly travis doesn't cover) to check it out.

Mind you I'm not saying all these things will *definitely* cause a problem - more that it's possible, and I want to both be confident in this PR's correctness and also avoid leaving behind a footgun for the future.

jdalton:
> Mind you I'm not saying all these things will definitely cause a problem - more that it's possible, and I want to both be confident in this PR's correctness and also avoid leaving behind a footgun for the future.

Cool. Ya, the use case I'm going for is to be a replacement for the linear search for object keys. That means it just needs to be a black box that is a yay or nay for if an object key is found. It can even go along with the array tracking as just a faster check but defer to the linear backing for when order is important.

jdalton:
Looks like unit tests pass in IE11 and it does [resolve the perf issue](https://github.com/angular/angular/issues/8112#issuecomment-261681855) :tada:

ljharb:
Thanks, I'll give it a thorough review, and do some extensive testing in as many browsers as I can before merging this.

jdalton:
Tested on Safari 7, 8, 9, 10 and IE 9, 10 with no Map related fails.

ljharb:
TC39 is this week so I'll need a bit more time to review; rest assured it's still on my list :-)

ljharb:
Thanks!

PR #436, opened 2017-03-01
Correct version numbers for .min files

Alex-Sokolov:


ljharb:
This seems like a fine PR.

PR #437, opened 2017-03-19
Distinguish Set and Map iterators, with tests

### LINKED ISSUE #387: Map/Set iterator

Xotic750:
Closes #387. Closes #391.

### DIFF
--- a/test/set.js
+++ b/test/set.js
@@ -45,6 +45,7 @@ describe('Set', function () {
   var functionsHaveNames = (function foo() {}).name === 'foo';
   var ifFunctionsHaveNamesIt = functionsHaveNames ? it : xit;
   var ifShimIt = (typeof process !== 'undefined' && process.env.NO_ES6_SHIM) ? it.skip : it;
+  var ifGetPrototypeOfIt = Object.getPrototypeOf ? it : xit;
 
   var range = function (from, to) {
     var result = [];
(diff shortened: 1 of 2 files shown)


test/set.js
```diff
@@ -634,4 +634,22 @@ describe('Set', function () {
 
   it.skip('should throw proper errors when user invokes methods with wrong types of receiver', function () {
   });
+
+  it('SetIterator identification', function () {
```
ljharb: we should conditionally skip this first test when `Object.getPrototypeOf` is not available (in a separate `it`; same in the Map tests)

Xotic750:
Were there any change requests that I missed, or is there anything more to add for this?

es6-shim.js
```diff
@@ -2842,7 +2842,11 @@
         };
 
         MapIterator.prototype = {
+          isMapIterator: true,
```
ljharb: it occurs to me that this is too easily forgeable; but that since `instanceof` can't work cross-realm, this might be the best we can get.

Xotic750: Can't think why anyone would want to forge it, but I also can't think of anything to make it more unforgeable.

ljharb:
For posterity; this PR included ba1770b3e0e862e92736419344da5dca1d8441e8, 79d80a61667dd423706bd1d36814a25aec02007c, and b1d9b8438cee953b4e96785cb3f97d34f0fd00ab; I mistakenly merged the first two and forgot the third.

Issue #459, opened 2019-09-27
Shimmed Symbol.iterator is not recognized


wintlu:
When we include a Symbol shim before es6-shim, `Symbol.iterator` key is not correctly recognized, and this caused bunched of other issues. (for example, cannot retrieve correct iterator from ArrayIterator object)

**To reproduce:**
1. open this link in IE11:
https://wintlu.github.io/es6-shim-symbol-repo.html  ([source code here](https://github.com/wintlu/wintlu.github.io/blob/master/es6-shim-symbol-repo.html))
2. Check error is thrown

**Solution:**
Check symbol value for native and shimmed respectively:

```javascript
      var isNativeSymbol = function isNativeSymbol(value) {
        return typeof globals.Symbol === 'function' && typeof x === 'symbol';
      }

      var isShimSymbol = function isShimSymbol(value) {
        return typeof globals.Symbol === 'function' && Object(x) instanceof globals.Symbol;
      }

      var Type = {
        symbol: function (x) {
          return isNativeSymbol(x) || isShimSymbol(x);
        }
      }
```
Refer to my sample PR before: https://github.com/wintlu/google-map-es6-shim/pull/1/files

Want to confirm if my understanding and solution is correct, 
as always, many thanks to @lencioni

ljharb:
Symbols can not be shimmed - only shammed. As such, I'd suggest not trying to include them.

If you do, you'd have to include a symbol sham *after* es6-shim.

ljharb:
Symbols can not be shimmed - only shammed. As such,

wintlu:
Hi @ljharb , I see you reopened this issue, so we will need more discussion here?
actually I solution I pasted is from core-js:
https://github.com/zloirock/core-js/blob/master/packages/core-js/modules/es.symbol.js#L88
So would it make sense to have such check?

ljharb:
I closed it by accident, is all.

No - it would be incorrect and inappropriate to use anything but typeof to check for symbols, imo. Additionally, instanceof doesn’t work across realms.

wintlu:
thanks Jordan for the explanation! I agree with your point here
Right now I am facing an issue that seems either need to be fixed in google maps API or es6-shim side, could you provide me some suggestions for the fix? @ljharb @lencioni

**Context:**
1. We need to upgrade to google map js API 3.37 or 3.38 by end of Oct. 2019, (it will auto update if we don't do anything), IE11 will break if we don't fix the issue
[...]

PR #460, opened 2019-10-16
[Fix] Add non-native symbol check

wintlu:
Fix issue #459 

When we include a Symbol shim before `es6-shim`, Symbol.iterator key is not correctly recognized, and this caused bunched of other issues. (for example, cannot retrieve correct iterator from ArrayIterator object)

### DIFF
--- a/es6-shim.js
+++ b/es6-shim.js
@@ -255,13 +255,22 @@
   };
   var isArguments = isStandardArguments(arguments) ? isStandardArguments : isLegacyArguments;
 
+  var hasNativeSymbol = typeof Symbol === 'function' && typeof Symbol() === 'symbol';
+
+  var isNativeSymbol = function isNativeSymbol(value) {
+    return hasNativeSymbol && typeof value === 'symbol';
+  };
+
+  // idea from https://github.com/zloirock/core-js/blob/master/packages/core-js/modules/es.symbol.js#L88
+  var isNonNativeSymbol = function isNonNativeSymbol(value) {
+    return typeof globals.Symbol === 'function' && Object(value) instanceof globals.Symbol;
+  };
+
   var Type = {
     primitive: function (x) { return x === null || (typeof x !== 'function' && typeof x !== 'object'); },
     string: function (x) { return _toString(x) === '[object String]'; },
     regex: function (x) { return _toString(x) === '[object RegExp]'; },
-    symbol: function (x) {
-      return typeof globals.Symbol === 'function' && typeof x === 'symbol';
-    }
+    symbol: hasNativeSymbol ? isNativeSymbol : isNonNativeSymbol
   };
 
   var overrideNative = function overrideNative(object, property, replacement) {
@@ -295,7 +304,7 @@
   var $String = String;
 
   /* global document */
-  var domAll = (typeof document === 'undefined' || !document) ? null : document.all;
+  var domAll = typeof document === 'undefined' || !document ? null : document.all;
   var isNullOrUndefined = domAll == null ? function isNullOrUndefined(x) {
     return x == null;
   } : function isNullOrUndefinedAndNotDocumentAll(x) {
--- a/es6-shim.js
+++ b/es6-shim.js
@@ -304,7 +304,7 @@
   var $String = String;
 
   /* global document */
-  var domAll = typeof document === 'undefined' || !document ? null : document.all;
+  var domAll = (typeof document === 'undefined' || !document) ? null : document.all;
   var isNullOrUndefined = domAll == null ? function isNullOrUndefined(x) {
     return x == null;
   } : function isNullOrUndefinedAndNotDocumentAll(x) {


es6-shim.js
```diff
@@ -255,12 +255,21 @@
...
+    return typeof globals.Symbol === 'function' && Object(value) instanceof globals.Symbol;
+  };
+
   var Type = {
     primitive: function (x) { return x === null || (typeof x !== 'function' && typeof x !== 'object'); },
     string: function (x) { return _toString(x) === '[object String]'; },
     regex: function (x) { return _toString(x) === '[object RegExp]'; },
     symbol: function (x) {
-      return typeof globals.Symbol === 'function' && typeof x === 'symbol';
+      return isNativeSymbol(x) || isNonNativeSymbol(x);
```
ljharb: i'd prefer to not even call `isNonNativeSymbol` in an environment with native symbols. Can we change this check so that `Type.symbol` itself is either `=== isNativeSymbol`, or `=== isNonNativeSymbol`, rather than having to do both checks at runtime?

es6-shim.map
```diff
@@ -1 +1 @@
...,KACxB,IAAIioB,IAAS,KAAM,CAEjBxB,EAAQlqB,KAAKyrB,SAASC,EACtB,IAAIxB,EAAO,CACT,MAAOA,GAAM7nB,UACR,CACL,QAGJ,GAAIrC,KAAKurB,KAAM,CAEbrB,EAAQniB,EAAW5H,KAAKH,KAAKurB,KAAM9nB,EACnC,IAAIymB,EAAO,CACT,MAAOA,GAAM7nB,UACR,CACL,QAGJ,GAAI6oB,GAAOlrB,KAAKmrB,KAChB,IAAI7Z,GAAI4Z,CACR,QAAQ5Z,EAAIA,EAAEvE,QAAUme,EAAM,CAC5B,GAAIzgB,GAAGqB,cAAcwF,EAAE7N,IAAKA,GAAM,CAChC,MAAO6N,GAAEjP,SAKf4F,IAAK,QAASA,KAAIxE,GAChBunB,EAAehrB,KAAM,MACrB,IAAI0rB,GAAO7B,GAAQpmB,EAAK,KACxB,IAAIioB,IAAS,KAAM,CAEjB,aAAc1rB,MAAKyrB,SAASC,KAAU,YAExC,GAAI1rB,KAAKurB,KAAM,CAEb,MAAOvjB,GAAW7H,KAAKH,KAAKurB,KAAM9nB,GAEpC,GAAIynB,GAAOlrB,KAAKmrB,KAChB,IAAI7Z,GAAI4Z,CACR,QAAQ5Z,EAAIA,EAAEvE,QAAUme,EAAM,CAC5B,GAAIzgB,GAAGqB,cAAcwF,EAAE7N,IAAKA,GAAM,CAChC,MAAO,OAGX,MAAO,QAGTK,IAAK,QAASA,KAAIL,EAAKpB,GACrB2oB,EAAehrB,KAAM,MACrB,IAAIkrB,GAAOlrB,KAAKmrB,KAChB,IAAI7Z,GAAI4Z,CACR,IAAIhB,EACJ,IAAIwB,GAAO7B,GAAQpmB,EAAK,KACxB,IAAIioB,IAAS,KAAM,CAEjB,SAAW1rB,MAAKyrB,SAASC,KAAU,YAAa,CAC9C1rB,KAAKyrB,SAASC,GAAMrpB,MAAQA,CAC5B,OAAOrC,UACF,CACLkqB,EAAQlqB,KAAKyrB,SAASC,GAAQ,GAAIf,GAASlnB,EAAKpB,EAChDiP,GAAI4Z,EAAKN,UAGN,IAAI5qB,KAAKurB,KAAM,CAEpB,GAAIvjB,EAAW7H,KAAKH,KAAKurB,KAAM9nB,GAAM,CACnCsE,EAAW5H,KAAKH,KAAKurB,KAAM9nB,GAAKpB,MAAQA,MACn
```
ljharb: please revert the built files; they're only updated by releases.

es6-shim.js
```diff
@@ -295,7 +304,7 @@
   var $String = String;
 
   /* global document */
-  var domAll = (typeof document === 'undefined' || !document) ? null : document.all;
+  var domAll = typeof document === 'undefined' || !document ? null : document.all;
```
wintlu: will remove these changes

wintlu:
hi @ljharb, need your help on writing tests (don't quite understand how tests work in this repo):
1. which file should this symbol test go? can I reuse array.js or do I need to create new file like `test/symbol.js`?
1. how do I run test? do I simply run it locally by `npm run test`, or will the CI run it in real IE11?
1. how do I write the test? Do I need to create a dummy Symbol and verify on it? something like:
```js
const Symbol = () => {}
Symbol.iterator = { next: () => {}, done: false };

it('array has non native Symbol.iterator', () => {
  expect([][Symbol.iterator]).to.NotNull()
});
```

ljharb:
It's probably better to make a new file, but it's fine to reuse the array file as well. I can reorganize once you've added the test, if needed :-)

To run the tests in IE 11, you'll have to load https://github.com/paulmillr/es6-shim/blob/master/test/index.html in IE 11 yourself, manually.

I'd expect that in an environment with native symbols, the test would just use them, but in an environment without them, it would replicate the conditions that cause #459 - such that the test fails prior to this fix.

wintlu:
google maps fixed in their code, so we don't need to make this change here, thanks!

PR #462, opened 2020-02-28
Don't mix formal parameters and arguments array

milk-z:
IE11 perf is significantly degraded when mixing formal parameters and the arguments array

Perf comparisons of mixing formal parameters and arguments vs not shows not mixing is much faster. And using only formal parameters vs using only arguments array shows similar perf.
mixed-perf
unmixed-perf

### DIFF
--- a/es6-shim.js
+++ b/es6-shim.js
@@ -1387,7 +1387,8 @@
     var isBadHex = isBadHexRegex.test.bind(isBadHexRegex);
     var NumberShim = (function () {
       // this is wrapped in an IIFE because of IE 6-8's wacky scoping issues with named function expressions.
-      var NumberShim = function Number(value) {
+      var NumberShim = function Number() {
+        var value = arguments[0];
         var primValue;
         if (arguments.length > 0) {
           primValue = Type.primitive(value) ? value : toPrimitive(value, 'number');


es6-shim.js
```diff
@@ -1387,7 +1387,8 @@
     var isBadHex = isBadHexRegex.test.bind(isBadHexRegex);
     var NumberShim = (function () {
       // this is wrapped in an IIFE because of IE 6-8's wacky scoping issues with named function expressions.
-      var NumberShim = function Number(value) {
```
ljharb: this would be a breaking change; `Number.length` must be `1`.

es6-shim.js
```diff
@@ -1387,7 +1387,8 @@
...
+      var NumberShim = function Number() {
+        var value = arguments[0];
```
ljharb: but i'm fine using the *value* out of the arguments object - however, if `arguments.length < 1`, `arguments[0]` deoptimizes in many engines, including older v8.

es6-shim.js
milk-z: I missed that requirement! Thanks.
I will need to think about this some more then, just by having the formal parameter, even if only arguments is used, IE11 is much slower than only having formal parameters or only having arguments.

ljharb: Unfortunately function lengths are not configurable in IE 11, so the only way to get a length of 1 is to have a formal parameter.

PR #465, opened 2021-03-11
upgrade jquery to avoid security vulnerabilities

annieredmond1:
This updates jquery to avoid this security vulnerability https://github.com/advisories/GHSA-jpcq-cgw6-v4j6

### DIFF
--- a/test-sham/index.html
+++ b/test-sham/index.html
@@ -3,7 +3,7 @@
 <head>
   <meta charset="utf-8">
   <title>es6-shim tests</title>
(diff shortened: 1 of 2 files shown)


ljharb:
I'm pretty sure we're not vulnerable to XSS from untrusted sources in our *test files*.

ljharb (changes requested):
Does jQuery v3.5.0 work in IE 6? If not, this is not a viable upgrade.

PR #466, opened 2021-04-28
Fix crash issue when using es5-sham 4.5.15

gnh1201:
https://github.com/es-shims/es5-shim/issues/476

### DIFF
--- a/es6-shim.js
+++ b/es6-shim.js
@@ -1268,7 +1268,10 @@
   // Chrome defines keys/values/entries on Array, but doesn't give us
   // any way to identify its iterator.  So add our own shimmed field.
   if (Object.getPrototypeOf) {
-    addIterator(Object.getPrototypeOf([].values()));
+    var ChromeArrayIterator = Object.getPrototypeOf([].values());
+    if (ChromeArrayIterator) { // in WSH, this is `undefined`
+      addIterator(ChromeArrayIterator);
+    }
   }
 
   // note: this is positioned here because it relies on Array#entries


es6-shim.js
```diff
@@ -69,6 +69,7 @@
   var _some = Function.call.bind(Array.prototype.some);
 
   var defineProperty = function (object, name, value, force) {
+    object = object || {};
     if (!force && name in object) { return; }
```
ljharb, line 73: reassigning arguments deoptimizes the function. instead:
```suggestion
    object = object || {};
    if (!object || (!force && name in object)) { return; }
```

altho i'd rather fix all the callers of `defineProperty` and ensure they're not passing a non-object here.

es6-shim.js
```diff
@@ -665,7 +666,7 @@
   var addIterator = function (prototype, impl) {
     var implementation = impl || function iterator() { return this; };
     defineProperty(prototype, $iterator$, implementation);
-    if (!prototype[$iterator$] && Type.symbol($iterator$)) {
+    if (prototype && !prototype[$iterator$] && Type.symbol($iterator$)) {
```
ljharb, line 669: similarly, this change is fine but when is `addIterator` called with a falsy `prototype` value??

ljharb (changes requested):

es6-shim.js
gnh1201: When the `object` variable in the `defineProperty` method is marked as `undefined`, the `prototype` is also marked as `undefined`.

ljharb: Why would addIterator ever get called in this case tho?

gnh1201: I captured `arguments.callee.caller` inside `addIterator` method to check this.
Most of the `addIterator` calls were made from top-level functions. The reason why `addIterator` is called seems to be because there are no more errors in `defineProperty`.

gnh1201: I don't think there's anything strange about `addIterator` being called.

ljharb: It is absolutely a bug in es6-shim if `addIterator` is called with non-object arguments. That should be fixed directly - this approach is just papering over the real problem.

gnh1201: I tracked the problem and found the following results.

```
StringIterator.prototype: object
String.prototype: object
ArrayIterator.prototype: object
Array.prototype: object
Object.getPrototypeOf([].values()): undefined
```

ljharb: What does `[].values.toString()` print out?

gnh1201:
```
* typeof([].values)
function

* [].values.toString()
function values() {
      return new ArrayIterator(this, 'value');
    }
```

ljharb: Hmm, I'm super confused. ArrayIterator is a normal function, and a new instance of it should just inherit from Object.prototype, which should itself inherit from null. What does `Object.getPrototypeOf(Object.prototype)` print out?

gnh1201:
```
* typeof(Object.getPrototypeOf(Object.prototype))
undefined

* Object.getPrototypeOf(Object.prototype)
[BLANK]
```

gnh1201: I think it's a problem because the engine doesn't have the `__proto__` keyword.

```
* typeof([].values().__proto__)
undefined

* ('__proto__' in [].values()).toString()
false
```

Maybe we need a polyfill for `__proto__`. Or I think we need to find an alternative way (because it is deprecated)

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/proto

ljharb: Ahhh - when `__proto__` doesn't exist, `Object.getPrototypeOf` probably shouldn't be shammed at all. However, there's supposed to be a special exception in the sham so that `Object.getPrototypeOf({})` returns `Object.prototype` - but it _doesn't_ have a special exception so that getting the proto of Object.prototype returns `null`. Let me think a bit about what to do here.

ljharb: if the only change made is to the `if` block on line 1270, to something like:
```js
if (Object.getPrototypeOf) {
  var ChromeArrayIterator = Object.getPrototypeOf([].values());
  if (ChromeArrayIterator) { // in WSH, this is `undefined`
    addIterator(ChromeArrayIterator);
  }
}
```

does that fix the issue?

gnh1201: Fixed! There are no errors.

ljharb: awesome! let's update this PR to make that change, and we can get this merged

gnh1201: Thanks :)

es6-shim.js
```diff
@@ -69,7 +69,7 @@
   var _some = Function.call.bind(Array.prototype.some);
 
   var defineProperty = function (object, name, value, force) {
-    if (!force && name in object) { return; }
+    if (!object || (!force && name in object)) { return; }
```
ljharb, line 72: do we need this change now? if we had to keep the object check, i'd prefer to make it throw, so the error can be surfaced quickly, but instead i'd rather count on always calling `defineProperty` correctly.

es6-shim.js
```diff
@@ -665,7 +665,7 @@
   var addIterator = function (prototype, impl) {
     var implementation = impl || function iterator() { return this; };
     defineProperty(prototype, $iterator$, implementation);
-    if (!prototype[$iterator$] && Type.symbol($iterator$)) {
+    if (prototype && !prototype[$iterator$] && Type.symbol($iterator$)) {
```
ljharb, line 668: similarly here, i'd rather throw an exception when prototype is falsy if we need the check; otherwise can we revert it?

es6-shim.js
gnh1201: I think it is okay to revert this.

es6-shim.js
gnh1201: I think it is okay to revert this.

ljharb (approved):

Issue #467, opened 2021-04-29
babel-polyfill and es6-shim


gnh1201:
## Related items
* babel-polyfill: https://cdnjs.com/libraries/babel-polyfill
* Pull request https://github.com/paulmillr/es6-shim/pull/466
* Issue https://github.com/es-shims/es5-shim/issues/476

## Description
To run ES5 or higher Javascript on WSH(Windows Scripting Host), we need a `babel-polyfill`.

In this case, it should be changed as below to works.

```javascript
  var defineProperty = function (object, name, value, force) {
    if (!object || (!force && name in object)) { return; }
```

This change has also been proposed in https://github.com/paulmillr/es6-shim/pull/466.

I think that this change should apply.

gnh1201:
This code works well with only `babel-polyfill` and es5 shim and sham. The problem occurs when adding `es6-shim` here.

ljharb:
Is this still an issue after #466 is merged?

gnh1201:
Yes. This seems to be another problem with #466.

gnh1201:
shimtest.js: https://gist.github.com/gnh1201/2627d860e41f8bd2dba858536836c64e

zloirock:
@gnh1201 why do you need both - `core-js` and `es-shims`? It's a bad practice. What does `es-shims` fix in this case? BTW `babel-polyfill` is obsolete for a long time ago and the actual version of `core-js` should be used directly.

gnh1201:
@zloirock I needed a single js file to proceed with polyfill. I recognized `core-js`, but it looked different from the way it was used in legacy. It is burdensome to do such work at WSH.
Eventually, I had no choice but to choose `babel-polyfill` and `es5-shim`. It worked well and successfully executed the JS library file written in es5.
Trying `es6-shim` was an option for me, but I thought I'd still test it.

ljharb:
While most of the es-shims should work fine alongside core-js, it's definitely odd to mix them.

I would expect that es5-shim and es6-shim alone could get you up to speed, except for the Symbol shams that core-js includes but es6-shim does not.

gnh1201:
I tested core-js and confirmed that the example code works well. But I still have to wait and see. I think I need to check the impact when es-shims is excluded from my project.

PR #470, opened 2021-04-30
Fix crash when `target`(in Value.preserveToString) is undefined

gnh1201:
https://github.com/paulmillr/es6-shim/issues/467

### DIFF
--- a/es6-shim.js
+++ b/es6-shim.js
@@ -137,7 +137,7 @@
       }
     },
     preserveToString: function (target, source) {
-      if (source && isCallable(source.toString)) {
+      if (target && source && isCallable(source.toString)) {
         defineProperty(target, 'toString', source.toString.bind(source), true);
       }
     }


gnh1201:
```
* typeof(globals.Reflect): object
* typeof(ReflectShims.defineProperty): undefined
overrideNative(globals.Reflect, 'defineProperty', ReflectShims.defineProperty); <-- error
```

This means that `globals.Reflect` variable is going to be `undefined`. 

How can I fix a problem that `ReflectShims.defineProperty` is `undefined`?

ljharb:
Ah - if `Reflect` exists but `Reflect.defineProperty` does not, then that block of code should do branch, and call `defineProperty` versus `overrideNative`.

gnh1201:
Found a better solution.

ljharb:
@gnh1201 what better solution? Is #467 still an issue?

PR #471, opened 2021-07-29
Fix `ReflectShims.defineProperty` is undefined in WSH

gnh1201:
* Reverted from #466 - No need to check that this is `undefined`.
* This pull request fixes the following issue: #470 #467

### DIFF
--- a/es6-shim.js
+++ b/es6-shim.js
@@ -1264,10 +1264,7 @@
   // Chrome defines keys/values/entries on Array, but doesn't give us
   // any way to identify its iterator.  So add our own shimmed field.
   if (Object.getPrototypeOf) {
-    var ChromeArrayIterator = Object.getPrototypeOf([].values());
-    if (ChromeArrayIterator) { // in WSH, this is `undefined`
-      addIterator(ChromeArrayIterator);
-    }
+    addIterator(Object.getPrototypeOf([].values());
   }
 
   // note: this is positioned here because it relies on Array#entries
@@ -3585,7 +3582,7 @@
     });
   }
 
-  if (supportsDescriptors) {
+  if (!ReflectShims.defineProperty) {    // possible alternative: if (supportsDescriptors)
     var internalGet = function get(target, key, receiver) {
       var desc = Object.getOwnPropertyDescriptor(target, key);
 


es6-shim.js
```diff
@@ -3585,7 +3582,7 @@
     });
   }
 
-  if (supportsDescriptors) {
+  if (!ReflectShims.defineProperty) {    // possible alternative: if (supportsDescriptors)
```
ljharb, line 3585: `supportsDescriptors` is only true when get/set are available - but `ReflectShims.defineProperty` might be available in the same way `Object.defineProperty` might be - working fine for enumerable/configurable/writable data properties, but throwing for everything else. What does this fix?

es6-shim.js
```diff
@@ -1264,10 +1264,7 @@
   // Chrome defines keys/values/entries on Array, but doesn't give us
   // any way to identify its iterator.  So add our own shimmed field.
   if (Object.getPrototypeOf) {
-    var ChromeArrayIterator = Object.getPrototypeOf([].values());
-    if (ChromeArrayIterator) { // in WSH, this is `undefined`
-      addIterator(ChromeArrayIterator);
-    }
```
ljharb, line 1270: is this no longer true? iow, in WSH, `Object.getPrototypeOf([].values());` is not actually `undefined`?

ljharb (changes requested):

es6-shim.js
gnh1201: This change is a key point to solve #466 and other `overrideNative` related crash issues.

In WSH(polyfilled by core-js), `supportsDescriptors`(es6-shim) is false bacause of `if(!supportsAccessors && (hasGetter || hasSetter))`(es5-shim) is true.

Because of this, `replacement` parameter is undefined when calling `overrideNative`. This causes the wrong override.

es6-shim.js
gnh1201: Only if you apply this change.

es6-shim.js
gnh1201: This part should be run regardless of the values in `supportsDescriptors`.

gnh1201:
Any update?

es6-shim.js
ljharb: I'm still a bit unclear on your explanation.

If you're in WSH, polyfilled by core-js, why are you running es5-shim and es6-shim at all? You're supposed to use core-js OR es-shims, not both.

gnh1201:
I'll open it again when the issue becomes clear.

ljharb:
I’d love to understand what’s going on - do issues appear when not using core-js?

PR #480, opened 2025-11-16
[Refactor] Harden against pathological function names

ericcornelissen:
Change the regular expression for matching the function name to avoid potential quadratic runtime of the implementation of `Function.prototype.name` due to the regular expression in `es6-sham.js:157`.

I'm not aware of any input that would trigger such behavior, however the fix in this case is trivial. In fact, both the leading `\s` and trailing `\s` in the regular expression are unnecessary (though, based on context, I would say the leading `\s` is actually necessary, however it currently isn't so...).

See also https://github.com/es-shims/Function.prototype.name/pull/37


ljharb:
If there's no regression test I'm not sure it's worth fixing.

ericcornelissen:
I'm not sure how to add tests because this branch

https://github.com/paulmillr/es6-shim/blob/65f0a3e573564d3e0b322e8516a75cccfc6a3e90/es6-sham.js#L147

is never taken. This is because `function foo() {}.name !== 'foo'` is always false, at least on any environment I can get my hands on. I tested various Node.js versions down to v0.8.6.

Instead, I'd like to argue that this change makes sense simply because it aligns the implementation with https://github.com/es-shims/Function.prototype.name/pull/37.

ljharb:
That will be false in IE 6-8.

ericcornelissen:
Added tests now. I validated them manually by forcing the branch in `es6-sham.js` with `true ||` and deleting the `name` property in `test-sham/function.js`.

Also, it's probably worth noting that I manually edited `es6-sham.min.js` because the [build command](https://github.com/paulmillr/es6-shim/blob/65f0a3e573564d3e0b322e8516a75cccfc6a3e90/package.json#L40) for it does not work on my end, failing with the following output

```log
> es6-shim@0.35.8 minify:sham
> uglifyjs es6-sham.js --support-ie8 --keep-fnames --comments --source-map=es6-sham.map -m -b ascii_only=true,beautify=false > es6-sham.min.js

node:internal/errors:546
      throw error;
      ^

TypeError [ERR_INVALID_ARG_TYPE]: The "data" argument must be of type string or an instance of Buffer, TypedArray, or DataView. Received an instance of Object
    at Object.writeFileSync (node:fs:2407:5)
    at /es6-shim/node_modules/uglify-js/bin/uglifyjs:485:16
    at /es6-shim/node_modules/uglify-js/node_modules/async/lib/async.js:188:33
    at /es6-shim/node_modules/uglify-js/bin/uglifyjs:389:9
    at FSReqCallback.readFileAfterClose [as oncomplete] (node:internal/fs/read/context:68:3) {
  code: 'ERR_INVALID_ARG_TYPE'
}

Node.js v25.9.0
```

test-sham/function.js
```diff
@@ -54,5 +69,26 @@ describe('Function', function () {
...
+
+    it('runs on time for functions with lots of whitespace', function () {
+      var func = eval('(function(){' + new Array(100001).join(' ') + '})'); // eslint-disable-line no-eval
+
+      var start = Date.now();
+      var result = func.name;
+      var elapsed = Date.now() - start;
+
+      expect(result).to.equal('');
+      expect(result).to.be.lessThan(250);
```
ericcornelissen, line 91: I added this explicit check even though [Mocha](https://github.com/paulmillr/es6-shim/blob/65f0a3e573564d3e0b322e8516a75cccfc6a3e90/package.json#L74) will complain if the test takes too long already, happy to remove or [replace with a native solution](https://mochajs.org/features/timeouts/) it if that's preferred.

